Issue #12, opened 2018-07-05
LCM Traffic Configuration


mposa:
Configure all LCM traffic to appropriately communicate either locally or remotely. Likely requires passing through some (but not all) of high-frequency traffic like state and input messages.

PR #20, opened 2018-08-17
Telemetry Reduction

mshalm:
Implementation of first fix for #12.

LCM traffic has been separated into two categories on the same group; traffic by default has time-to-live 0 so that it does not make it to the router from cassie. Telemetry data (as well as procman-deputy) is elevated to TTL 1. Tested on cassie by passively moving the motors. Telemetry is broadcast on the CASSIE_TELEMETRY channel


mposa:
Superseded by #40

PR #34, opened 2019-02-18
Cassie sensor simulation

yminchen:
Add simulated cassie sensors (only for RigidbodyPlant).

PR #41, opened 2019-02-21
Cassie sensor simulation (continued)

yminchen:
Continuing the PR #34


examples/Cassie/cassie_utils.h
```diff
@@ -46,6 +62,32 @@ void buildCassieTree(
     drake::multibody::joints::FloatingBaseType base_type =
         drake::multibody::joints::kFixed);
 
+/// Add simulated accelerometer to the diagram
+drake::systems::sensors::Accelerometer * addSimAccelerometer(
```
mposa: Indentation, here and below

examples/Cassie/run_simple_sim_with_cassie_sensor.cc
```diff
@@ -0,0 +1,237 @@
...
+
+  drake::lcm::DrakeLcm lcm;
+  std::unique_ptr<RigidBodyTree<double>> tree;
+  if (FLAGS_floating_base)
+    tree = makeCassieTreePointer("examples/Cassie/urdf/cassie_v2.urdf",
+                                 drake::multibody::joints::kQuaternion);
+  else
+    tree = makeCassieTreePointer();
+
+  std::shared_ptr<RigidBodyFrame<double>> imu_frame =
```
mposa: Question--is the IMU located at the origin of the pelvis?

examples/Cassie/run_simple_sim_with_cassie_sensor.cc
```diff
@@ -0,0 +1,237 @@
+#include <memory>
```
mposa: Question--can this replace the basic simulation, instead of being added as a separate file? Would there be a downside to that?

systems/sensors/sim_cassie_sensor_aggregator.cc
```diff
@@ -0,0 +1,145 @@
...
+
+  positionIndexMap_ = multibody::makeNameToPositionsMap(tree);
+  velocityIndexMap_ = multibody::makeNameToVelocitiesMap(tree);
+  actuatorIndexMap_ = multibody::makeNameToActuatorsMap(tree);
+
+  input_input_port_ = this->DeclareVectorInputPort(
+                        BasicVector<double>(10)).get_index();
+  state_input_port_ = this->DeclareVectorInputPort(
+            BasicVector<double>(num_positions_ + num_velocities_)).get_index();
+  acce_input_port_ = this->DeclareVectorInputPort(
```
mposa: Why `acce` and not `accel`?

systems/sensors/test/sim_cassie_sensor_aggregator_test.cc
```diff
@@ -0,0 +1,56 @@
+#include "systems/sensors/sim_cassie_sensor_aggregator.h"
```
mposa: Is there an easy way to test that the results come back OK? Maybe confirm that the acceleration is 0 for the fixed base, and -g for the floating bases, and a similar test of non-zero angular velocity for the floating base models.

mposa:
Overall--looks great! Just a few questions/comments.

examples/Cassie/cassie_utils.cc
```diff
@@ -126,6 +126,65 @@ void buildCassieTree(RigidBodyTree<double>& tree, std::string filename,
   ankle_spring_joint_right.SetSpringDynamics(1250.0, 0.0);  // 2300 in URDF
 }
 
+drake::systems::sensors::Accelerometer * addSimAccelerometer(
+  drake::systems::DiagramBuilder<double> & builder,
+  drake::systems::RigidBodyPlant<double> * plant,
+  std::shared_ptr<RigidBodyFrame<double>> imu_frame){
+
+  auto acce_sim = builder.AddSystem<drake::systems::sensors::Accelerometer>(
+    "Simulated Accelerometer", *imu_frame, plant->get_rigid_body_tree(), true);
```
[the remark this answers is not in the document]
yminchen: Got it. I'm wondering what's the SublimeLinter setting for this double indent check? I'm also wondering if there is a way to automatically indent (more than just once) for this case.

examples/Cassie/cassie_utils.cc
```diff
@@ -126,6 +126,65 @@ void buildCassieTree(RigidBodyTree<double>& tree, std::string filename,
   ankle_spring_joint_right.SetSpringDynamics(1250.0, 0.0);  // 2300 in URDF
 }
 
+drake::systems::sensors::Accelerometer * addSimAccelerometer(
```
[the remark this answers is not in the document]
yminchen: Right. I'm passing the frame name (string) instead of RigidBodyFrame now.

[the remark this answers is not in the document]
mposa: Why would the user specify the frame name? That's true for some general robot, but this utilty is Cassie-specific.

examples/Cassie/cassie_utils.cc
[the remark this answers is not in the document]
mposa: I'm not aware of any, but let us all know if you find a good way to do it.

examples/Cassie/run_simple_sim_with_cassie_sensor.cc
yminchen: I replaced the basic simulation with my simulation with sensor. 
I prefer combining them into one file, so that we could reduce repeated code. I guess the downside is just that run_simple_sim.cc is now slight longer than before, and it puts more burden on the people who read it for the first time (but I think it's fine).

examples/Cassie/run_simple_sim_with_cassie_sensor.cc
yminchen: I'm not sure and I couldn't find it on the wiki. Just sent an email to Agility Robotics.

examples/Cassie/cassie_utils.cc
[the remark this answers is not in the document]
yminchen: No problem!

examples/Cassie/cassie_utils.cc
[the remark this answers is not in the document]
yminchen: I was thinking that since the frame name is declared in run_simple_sim (https://github.com/DAIRLab/dairlib/blob/2e1d1f64029a194c260cd43f22c7ff4f97f96c53/examples/Cassie/run_simple_sim.cc#L70),  I should pass in the name when I call the function. 

But on second thought, I agree that users shouldn't touch the frame name. I hid the name in cassie_utils just now.

systems/sensors/test/sim_cassie_sensor_aggregator_test.cc
yminchen: Do we need to test the IMU value? The job of SimCassieSensorAggregator is just gathering the data from accelerometer and gyroscope and put them into lcmt_cassie_out.

mposa: Fair enough.

examples/Cassie/run_simple_sim_with_cassie_sensor.cc
yminchen: Confirmed with Agility Robotics. The position in MuJoCo is correct. Also, the numbers are also available in Simulink simulator. 

e.g. https://github.com/osudrl/cassie-mujoco-sim/blob/master/model/cassie.xml#L86

PR #43, opened 2019-02-26
Multibody solvers

shrenikm:
# Constraint and solver classes for position, contact and fixed point solving.

## Changes to `ContactToolkit`
* The `xB` parameter has been removed from `ContactInfo`. All contacts represented through this structure are assumed to be point-plane contacts (Specifically the ground plane -- z plane).
* Added a new `num_contacts` (To store the number of contacts) to the structure to avoid having to use `idxA.size()` everywhere (Which was unintuitive).
* Default and parameterized constructors provided for `ContactInfo` to make object creation and setting default values easier.
* `CalcContactJacobian` has been updated to compute the Jacobian assuming point-ground plane contact. `xB` is no longer used.

## Constraint classes

The following constraint classes have been added:

### `PositionConstraint`
* Solves for tree position constraints. It is initialized using a `RigidBodyTree`.
### `ContactConstraint`
* Solves for contact constraints (Point-ground plane contacts). Object creation requires a `RigidBodyTree` and `ContactInfo` object.
### `FixedPointConstraint`
* Solves for fixed point constraints. Object creation requires a `RigidBodyTree` object and optionally a `ContactInfo` object.
* If a `ContactInfo` object is not provided, the fixed point is solved for the case with no contacts.
* if a `ContactInfo` object is provided, the contact constraints are also solved for. The contact forces are also taken into account while computing `xdot` in the constraint.

## Solver classes

The solver classes create objects of the Constraint classes and creates programs to solve them. They provide interfaces to set the initial values, obtain the program and results, verify constraints, set the output log file, constraint feasibility, etc. The code has been updated to comply with Drake's change to use `MathematicalProgramResult`. The following solver classes have been added:

### `PositionSolver`
* Creates a `PositionConstraint` object and solves the program.
* Contains a function to add joint limit constraints.
### `ContactSolver`
* Creates a `ContactConstraint` object and solves the contact program.
* Contains a function to add joint limit constraints.
### `FixedPointSolver`
* Creates a `FixedPointConstraint` object and solves the fixed point program. Depending on the constructor used, contact information may or may not be provided and the solver will solve the program accordingly.
* Contains functions to add a normal force spreading cost, friction cone constraints and joint limit constraints.

## Tests
* `contact_toolkit_test.cc` has been modified a bit to comply with the changes.
* `multibody_solvers_test.cc` contains the tests for the constraint and solver classes. All the classes and functions have been tested for the fixed and floating base Cassie models.

## `Multibody`
* The solver code has been placed in `attic/multibody/` together with the RBT version of `contact_toolkit`. `Multibody` versions of the solvers have not been written.
* `TreeConstraint` in `cassie_utils` has been removed and the dependency in `run_simple_sim` has been replaced with the `PositionSolver` version.


attic/multibody/multibody_solvers.cc
```diff
@@ -0,0 +1,568 @@
...
+  VectorXd joint_max = tree_.joint_limit_max;
+
+  for (int i = 0; i < joint_min.size(); ++i) {
+    // Adding minimum and maximum joint angle constraints.
+    prog_->AddConstraint(q_(i) >= (joint_min(i) + tolerance));
+    prog_->AddConstraint(q_(i) <= (joint_max(i) - tolerance));
+  }
+}
+
+MathematicalProgramResult ContactSolver::Solve(VectorXd q,
```
mposa: Recommend renaming `Solve`, since it modifies the problem by fixing some decision variables. Alternatively, split into two functions.

attic/multibody/test/contact_toolkit_test.cc
```diff
@@ -121,38 +123,49 @@ class ContactToolkitTest : public ::testing::Test {
...
 TEST_F(ContactToolkitTest, InitializationTest) {
+  // ContactInfo default values
+  ContactInfo default_contact;
+  ASSERT_TRUE(default_contact.xA.isApprox(Matrix3Xd::Zero(3, 1)));
+  ASSERT_EQ(default_contact.num_contacts, 0);
+
+  // ContactInfo construction validity check
+  Matrix3Xd xa = Matrix3Xd::Zero(3, 2);
+  vector<int> idxa(2, 1);
+  ContactInfo tmp_info = {xa, idxa};
```
mposa: Can a more descriptive name than tmp_info be used?

attic/multibody/contact_toolkit.cc
```diff
@@ -104,7 +104,7 @@ VectorX<T> ContactToolkit<T>::CalcMVDot(VectorX<T> x, VectorX<T> u,
   VectorX<T> right_hand_side =
       -tree_.dynamicsBiasTerm(k_cache, no_external_wrenches);
 
-  if (num_efforts > 0) {
+  if (num_efforts) {
```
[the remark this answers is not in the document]
shrenikm: Just thought it would be cleaner this way. Should I revert it back?

attic/multibody/test/contact_toolkit_test.cc
shrenikm: Yeah will do.

attic/multibody/multibody_solvers.cc
shrenikm: The `fixed_joints` parameter is optional though, so renaming it to reflect this could possibly be misleading. Do I go ahead and rename it?

mposa: What do you think of splitting it into two? I just don't like the idea of the Solve method doing anything more than solving the program. It would make it hard, for instance, to re-run the problem.

attic/multibody/contact_toolkit.cc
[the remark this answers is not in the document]
mposa: I think so. `>0` clearly describes what the check is for.

attic/multibody/multibody_solvers.cc
shrenikm: Even if it is split, we would end up with overlapping code. Can it be renamed to something like `FindSolution`?

attic/multibody/contact_toolkit.cc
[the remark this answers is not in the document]
shrenikm: Alright will do.

attic/multibody/multibody_solvers.cc
mposa: Ideally, we create optimization problems and then solve them, with the creation and the solving being two distinct and separable steps.

shrenikm: I see. So something like `FormulateProblem` and then `Solve`?

mposa: I'd recommend pulling out the setting of fixed joints into it's own method. The other parts related to setting up the `MathematicalProgram` could be done at construction of the `PositionSovler` (and other classes), correct?

shrenikm: Yeah that can be done. We'll also need to provide the values of `q`, etc. (The ones used in the cost) to the constructor in that case. Or would it be better to do everything apart from setting the cost and solving in the constructor? The `Solve` function would then set up the cost and solve the program.

mposa: If it makes sense, problem setup methods can be lumped together to simplify things from the user's perspective. The only thing I feel strongly about is separating out the solve from the setup.

shrenikm: 3c3fff9 tries to resolve this. I'm not sure if its the best way of doing it.

shrenikm:
Major updates: 
* Fixed joint constraints have been moved into a separate function. They are constructed using an `std::map`.
* The program setup and solve have been separated. The `Solve` function now only solves the program and updates the `MathematicalProgramResult` class member.
* The program setup is now handled by the respective constructor.
* Each solver class is provided with a function to add the program cost. Overloaded constructors are provided to enable program creation without the default cost or with a specific quadratic cost formulation. In the former case, the user must add the cost by calling the function before solving.
* All conflicts with master have been resolved.

attic/multibody/multibody_solvers.cc
```diff
@@ -0,0 +1,835 @@
...
+}
+
+PositionSolver::PositionSolver(const RigidBodyTree<double>& tree,
+                               VectorXd q_desired, MatrixXd Q)
+    : tree_(tree), prog_(make_shared<MathematicalProgram>()) {
+  // Checking the validity of q_desired
+  DRAKE_DEMAND(q_desired.size() == tree_.get_num_positions());
+
+  // If the user does not provide Q, default it to identity
+  if (Q.rows() == 0 && Q.cols() == 0) {
```
mposa: We shouldn't be passing in arguments that we know are wrong. I see why you did this, because you can't set a default parameter without knowing the dimension, but this isn't right. Here's what I recommend for the constructors:

```
 // default constructor, all other constructors delegate here
PositionSolver(tree) {
// basic program setup
}

PositionSolver(tree, q, Q) : PositionSolver(tree) {
  //Calls default constructor then adds cost
  AddProgramCost(q,Q);
}

PositionSolver(tree,q) : PositionSolver(tree,q,Identity(q.rows())) {}

mposa: This way, there's no duplicated code between constructors, and no strange if statements. Also, `q` and `Q` can and should be made `const` this way.

shrenikm: I see. The matrix check did seem to be a hacky way of doing it. I'll update it.

shrenikm: I guess I can also remove the default Q value from `AddProgramCost`. If the user calls it specifically, it's probably to change it from something other than identity.

shrenikm: dd09d84 implements this. 
* The default values are initialized through the constructor.
* Constructor code is reused.
* Default `Q` and `U` values from `AddProgramCost` have been removed.

PR #44, opened 2019-03-03
Run Drake Cassie sim with Dispatcher

### LINKED ISSUE #4: Build simulator that runs through the dispatcher
[run_simple_sim](https://github.com/DAIRLab/dairlib/blob/master/examples/Cassie/run_simple_sim.cc) is executes a simulation that listens to CASSIE_INPUT and outputs a CASSIE_STATE. This should be replaced by a simulator that emulates the actual robot, communicating via the Agility UDP API to the Dispatcher.

mposa:
-Update dispatcher to use simulated sensors.
-Added procman script to run Cassie PD demo with Drake simulation and dispatcher
-Removed default use of logfiles from multibody solvers, to avoid generating unused logs
-Reverted integration scheme for run_simple_sim to default value

Builds off of #41 to resolve #4

### DIFF
--- a/examples/Cassie/cassie_test.pmd
+++ b/examples/Cassie/cassie_test.pmd
@@ -14,8 +14,12 @@ group "1.simulated-robot" {
         exec = "bazel-bin/examples/Cassie/run_pd_controller";
         host = "localhost";
     }
-    cmd "2.dispatcher-robot-out" {
-        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001";
+    cmd "2.a.dispatcher-robot-out (drake)" {
+        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001 --simulation=true";
+        host = "localhost";
+    }
+    cmd "2.b.dispatcher-robot-out (gazebo)" {
+        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001 --simulation=true";
         host = "localhost";
     }
     cmd "1.simulator" {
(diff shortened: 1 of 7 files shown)


examples/Cassie/dispatcher_robot_out.cc
```diff
@@ -91,15 +110,27 @@ int do_main(int argc, char* argv[]) {
...
   auto diagram = builder.Build();
 
-  systems::UDPDrivenLoop loop(*diagram, *input_sub, nullptr);
+  if (FLAGS_simulation) {
+    drake::systems::lcm::LcmDrivenLoop loop(*diagram, *sim_input_sub, nullptr,
+        &lcm_local,
+        std::make_unique<UtimeMessageToSeconds<dairlib::lcmt_cassie_out>>());
+
+    // caused an extra publish call?
+    loop.set_publish_on_every_received_message(true);
```
yminchen: I forgot what you mentioned last time about the publish issue here. I think lcm driven loop forces LcmPublisherSystem to publish when a message arrives in LcmSubscriberSystem. When I set `set_publish_on_every_received_message` to false, the publish rate is 1k hz as expected. And when I set `set_publish_on_every_received_message` to true, the publish rate increased to 2k hz.

yminchen: I was just wondering if the problem causes us trouble. Maybe it's a bug in drake, but it doesn't matter for the users like us, as long as it publishes at the rate we want.

examples/Cassie/cassie_test.pmd
```diff
@@ -14,8 +14,12 @@ group "1.simulated-robot" {
...
         host = "localhost";
     }
-    cmd "2.dispatcher-robot-out" {
-        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001";
+    cmd "2.a.dispatcher-robot-out (drake)" {
+        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001 --simulation=true";
+        host = "localhost";
+    }
+    cmd "2.b.dispatcher-robot-out (gazebo)" {
+        exec = "bazel-bin/examples/Cassie/dispatcher_robot_out --port 25001 --simulation=true";
```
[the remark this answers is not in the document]
yminchen: Ah that's a typo. For gazebo simulation, the simulation flag should be set to false.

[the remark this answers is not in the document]
mposa: Good catch, thanks.

examples/Cassie/dispatcher_robot_out.cc
mposa: Good catch--the upstream issue has actually been fixed. The problem here was more mundane--both the simulator and dispatcher were publishing. Fixed with latest commit.

PR #50, opened 2019-03-15
Add a unit test file for dircon_util

yminchen:
- Added a simple example to test secondOrderCost() and linearizeConstraints(). 
- Shortened each line in dircon_util to less than 80 characters.


systems/trajectory_optimization/dircon_util.cc
```diff
@@ -144,49 +151,52 @@ VectorXd NVec(int start, int length) {
 }
 
 template <typename Derived>
-std::pair<int,int> getConstraintStart(const MathematicalProgram* prog, const std::vector<Binding<Derived>>& constraints,
-  Binding<Constraint>& c) {
+std::pair<int, int> getConstraintStart(const MathematicalProgram* prog,
+                               const std::vector<Binding<Derived>>& constraints,
```
mposa: Spacing here and below is a little wonky.

systems/trajectory_optimization/dircon_util.cc
```diff
@@ -24,8 +24,9 @@ void checkConstraints(const MathematicalProgram* prog,
     bool isSatisfied = (y.array() >= c->lower_bound().array() - tol).all() &&
                        (y.array() <= c->upper_bound().array() + tol).all();
     if (!isSatisfied) {
-      std::cout << "Constraint violation: " << c->get_description() << std::endl;
-      MatrixXd tmp(y.size(),3);
+      std::cout << "Constraint violation: "
+          << c->get_description() << std::endl;
```
mposa: I **think** the `<<` should go at the end of the line prior. It's easier to read that way.

systems/trajectory_optimization/dircon_util.cc
```diff
@@ -144,49 +151,52 @@ VectorXd NVec(int start, int length) {
...
-  Binding<Constraint>& c) {
+std::pair<int, int> getConstraintStart(const MathematicalProgram* prog,
+                               const std::vector<Binding<Derived>>& constraints,
+                               Binding<Constraint>& c) {
   int start = -1;
   int n = 0;
   for (auto const& binding : constraints) {
-    if (c.evaluator() == binding.evaluator() && c.variables() == binding.variables())
+    if (c.evaluator() == binding.evaluator()
+        && c.variables() == binding.variables())
```
mposa: Same as before. The `&&` should be on the prior line.

systems/trajectory_optimization/test/cost_constraint_approximation_test.cc
```diff
@@ -0,0 +1,114 @@
...
+                                 w);
+    // Adds a cost term of the form 0.5*x'*H_o*x + b_o'x + c
+    // intentionally split the following cost into two cost bindings
+    quadprog.AddQuadraticCost(H_o_ * 1 / 3, b_o_ * 1 / 3, w);
+    quadprog.AddQuadraticCost(H_o_ * 2 / 3, b_o_ * 2 / 3, w);
+    const auto result = Solve(quadprog);
+    w_sol_ = result.GetSolution(quadprog.decision_variables());
+
+    // Approximation
+    dairlib::systems::trajectory_optimization::linearizeConstraints(
```
mposa: The functions you're testing should probably be in the test itself, not the setup. Setup is largely useful for any shared constructors/problem setup that will be used across multiple tests without changing.

systems/trajectory_optimization/test/cost_constraint_approximation_test.cc
```diff
@@ -0,0 +1,114 @@
...
+        &quadprog, w_sol_, y_a_, A_a_, lb_a_, ub_a_);
+    double c_doub = dairlib::systems::trajectory_optimization::secondOrderCost(
+                      &quadprog, w_sol_, H_a_, b_a_);
+    c_a_ << c_doub;
+  }
+
+  MatrixXd H_o_ = MatrixXd::Zero(2, 2);
```
mposa: Are these `Zero` intializers necessary? Everything should be set before being read.

systems/trajectory_optimization/test/cost_constraint_approximation_test.cc
mposa: Also, there's nothing special about using `w_sol_`, any `w` could be used for this test.

yminchen: > The functions you're testing should probably be in the test itself, not the setup. Setup is largely useful for any shared constructors/problem setup that will be used across multiple tests without changing.

Got it!

yminchen: > Also, there's nothing special about using `w_sol_`, any `w` could be used for this test.

I used w_sol because w is the decision variable

systems/trajectory_optimization/test/cost_constraint_approximation_test.cc
yminchen: I think they are necessary. It gave segmentation fault when I didn't add them, so it seems that we need to pre-specify how large the size of the matrix/vector is.

yminchen: But this dosen't matter anymore, cause we are moving them out from setup()

yminchen:
Updated the test file. We don't prespecify the size of the matrices and vectors that we pass into the util functions.

systems/trajectory_optimization/test/cost_constraint_approximation_test.cc
```diff
@@ -0,0 +1,114 @@
...
+
+  VectorXd original_cost_value;
+  VectorXd approx_cost_value;
+  for (int i = 0; i < 3; i++) {
+    original_cost_value = 0.5 * w_sample[i].transpose() * H_o * w_sample[i] +
+                          b_o.transpose() * w_sample[i] + c_o;
+    approx_cost_value =
+        0.5 * (w_sample[i] - w_sol).transpose() * H_a * (w_sample[i] - w_sol)
+        + b_a.transpose() * (w_sample[i] - w_sol) + c_a;
+    ASSERT_TRUE( (original_cost_value - approx_cost_value).norm() < 1e-4 );
```
mposa: Couple of very minor things. Use 
```
#include "drake/common/test_utilities/eigen_matrix_compare.h"
EXPECT_TRUE(CompareMatrices(x, y, tol))
```
as shorthand for this. 

Also, here and below, `EXPECT_` should be used instead of `ASSERT_` unless failure of the condition renders the program unable to continue.

yminchen: Got it, thanks!

yminchen:
Minor updates according to the change request.

PR #51, opened 2019-03-16
Implement the AutoDiffXd version of HybridDircon

mposa:
While we can push this, I highly recommend not using it. It's slow and, for this example, produces poor quality results. Probably better to push it to master than leave unfinished code, but this requires some serious investigation once upstream https://github.com/RobotLocomotion/drake/issues/10322 is resolved.


mposa:
There's no hurry on a review here.

systems/trajectory_optimization/dircon_kinematic_data_set.h
```diff
@@ -59,25 +59,49 @@ class DirconKinematicDataSet {
...
+    }
+
+    std::size_t operator()(const CacheKey<drake::AutoDiffXd>& key) const {
+      using std::hash;
+      std::size_t ret = 0;
+      for (int i = 0; i < key.state.rows(); i++) {
+        ret = (ret << 1) ^ hash<double>{}(key.state(i).value());
+        for (int j = 0; j < key.state(i).derivatives().size(); j++) {
+          ret = (ret << 1) ^ hash<double>{}(key.state(i).derivatives()[j]);
+        }
```
yminchen: Just one comment -- I think this might not be necessary, since we probably don't get states with the same values but different derivatives? (same comment for input and force)

yminchen: I take the above comment back. Apparently it's necessary to encode the derivatives into the key, since SNOPT failed to find a solution after I removed it.

yminchen:
Looks good to me!

PR #53, opened 2019-03-16
Reorganized and cleaned up dircon_utils.

mposa:
-Moved to solvers/ directory as optimization_utils, since it is not dircon specific.
-Added documentation.
-Made use of new MathematicalProgram.GetAllConstraints().
-Changed all APIs to use better naming schemes and appropriate use of const references and pointers.


mposa:
No hurry on this one either. It's an API change, which will definitely effect your code. 

I'd appreciate a review of the API and documentation.

solvers/optimization_utils.cc
```diff
@@ -0,0 +1,175 @@
...
+          c->get_description() << std::endl;
+      MatrixXd tmp(y.size(), 3);
+      tmp << c->lower_bound(), y, c->upper_bound();
+      std::cout << tmp << std::endl;
+    }
+  }
+}
+
+double SecondOrderCost(const MathematicalProgram& prog, const VectorXd& x_nom,
+    MatrixXd* Q, VectorXd* w, double eps) {
```
yminchen: Just curious, why do we change from reference to pointer?

mposa: See the GSG here: https://google.github.io/styleguide/cppguide.html#Reference_Arguments

The main argument is that non-const references are confusing,making it too easy to unwittingly change data. Using pointers and const references makes this more explicit.

solvers/optimization_utils.h
```diff
@@ -0,0 +1,60 @@
...
+#include "drake/math/autodiff_gradient.h"
+
+namespace dairlib {
+namespace solvers {
+
+/// Check if the generic (nonlinear) constraints are satisfied up to the
+/// specified tolerance.
+/// Prints out any violating constraints
+/// @return True if all generic constraints are satisfied
+bool CheckConstraints(const drake::solvers::MathematicalProgram& prog,
```
yminchen: I was merging my branch with this branch, and I found that this seems to be a bug. Could either change this to `void CheckGenericConstraints` or modify the cc file so that the function returns a boolean variable.

solvers/optimization_utils.cc
yminchen: Got it!

solvers/optimization_utils.h
```diff
@@ -0,0 +1,60 @@
+#include "drake/solvers/mathematical_program.h"
```
yminchen: Sorry that I keep adding new comments after the approval. 
I think we need `#pragma once` here. Otherwise, we would re-define the default parameters such as eps in SecondOrderCost, and this would give compile error.

mposa: Done.

solvers/optimization_utils.h
mposa: Good catch. Changed to `bool CheckGenericConstraints`, does that work for you?

yminchen: I only spotted this bug and actually do not have code to test it. By reading the updated code, I think it's correct!

PR #54, opened 2019-03-20
LcmSubscriberMultiplexer and LcmMultiplexerDrivenLoop

mposa:
This PR adds two classes:

**LcmSubscriberMultiplexer**
A variation on `LcmSubscriberSystem`, this class has a factory method which actually creates `N` different copies of a `LcmSubscriberSystem`, on different channels, and connects the subscriber outputs to the Multiplexer's inputs. The Multiplexer has one additional input, a string, which governs the currently active subscriber. The output of the Multiplexer is a passthrough of the currently active Subscriber.

**LcmMultiplexerDrivenLoop**
A variation on `LcmDrivenLoop` where the active subscriber from a `LcmSubscriberMultiplexer` is used to drive the loop. This means that the driving subscriber can change. To handle cases where the driving subscriber stops publishing (e.g. due to a program crash), the `WaitForMessage` call is run with a timeout. 

This timeout is made possible by https://github.com/RobotLocomotion/drake/pull/10871


systems/lcm/lcm_multiplexer_driven_loop.cc
```diff
@@ -0,0 +1,103 @@
...
+  while (active_sub->WaitForMessage(message_count, nullptr, timeout_seconds_) <=
+      message_count) {
+    // Step to current time to update the input port, if it has changed
+    stepper_->StepTo(stepper_->get_context().get_time());
+
+    active_sub = driving_mux_.get_active_subscriber(context);
+  }
+  active_sub->CalcNextUpdateTime(*sub_context_, sub_events_.get());
+
+  // If active_sub->WaitForMessage() returned, a message should be received
```
yminchen: I took a second look, and I think this comment is not correct. active_sub->WaitForMessage() always return when we set a timeout. Maybe change to `If active_sub->WaitForMessage() > message_count, ...`?

systems/lcm/lcm_subscriber_multiplexer.cc
```diff
@@ -0,0 +1,109 @@
...
+using drake::systems::DiagramBuilder;
+using drake::systems::InputPort;
+using drake::systems::lcm::LcmSubscriberSystem;
+using drake::systems::lcm::SerializerInterface;
+using drake::systems::UnrestrictedUpdateEvent;
+using drake::systems::CompositeEventCollection;
+using drake::systems::State;
+
+namespace {
+const int kActiveChannelIndex = 0;
```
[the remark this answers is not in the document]
mposa: Changed to `kChannelSelectorPortIndex` to try to convey the idea but limit length (somewhat).

systems/lcm/lcm_multiplexer_driven_loop.cc
mposa: Done.

PR #55, opened 2019-03-21
Constrained LQR controller

shrenikm:
Implementation of a Constrained LQR controller and other small fixes.

## Controller Class

The controller is implemented through the `ConstrainedLQRController` class which is a subclass of `LeafSystem`. Although the controller implementation is affine, it does not extend `AffineController` due to difficulties in incorporating the required ports.

`constrained_lqr_controller_test.cc` is the unit test script for the controller class.

`run_constrained_lqr_controller.cc` is a run script that implements the controller for the fixed and floating base Cassie versions. The base type is given through the gflag `floating_base`. Controller gains have been tuned for the floating base implementation.

## RBT Visualizer

`rbt_visualizer.cc` has been added for floating base simulation (Which is not currently supported by `MultibodyPlant`). The corresponding entries have also been added to the procman script.

## Other fixes

* Fixed a bug in `AffineController` (As pointed out by @Nanda-Kishore-V ) in one of the setter methods. Added additional tests to the `AffineController` test script to account for this and a few other methods that weren't tested.
* Fixed a bug in `FixedPointSolver` where `AddSpreadingNormalForces` did not add the expression constraint after constructing the expression.
* Changed incorrect namespaces in some of the previous test files in `multibody_sovlers` and `contact_toolkit`.


examples/Cassie/run_constrained_lqr_controller_sim.cc
```diff
@@ -0,0 +1,472 @@
...
+}
+
+/*
+ * A class to serve as the input to the ConstrainedLQRController.
+ * It takes in a basic vector (of size num_states) at its input and has an
+ * output port of type OutputVector.
+ * It serves as an interfacing connector between the RigidBodyPlant's state
+ * output port and ConstrainedLQRController's info input port.
+ */
+class StateConnector : public LeafSystem<double> {
```
mposa: It feels like this should be moved to `multibody` and given a more appropriate name. It's acting like a state estimator from a simulated plant, but not currently copying over the efforts. I think it's OK, for now, to avoid adding in the effort logic (which will require an additional input port), but others would have use for a block like this.

examples/Cassie/run_constrained_lqr_controller_sim.cc
```diff
@@ -0,0 +1,472 @@
...
+                  input_pub->get_input_port());
+
+  // Constrained LQR controller
+
+  // Defining the Q and R matrix
+  // Positions for which the cost need to be ignored.
+  vector<string> cost_ignore_position;
+  if (FLAGS_floating_base) {
+    cost_ignore_position = {"base_x",    "base_y",     "base_z",
+                            "base_roll", "base_pitch", "base_yaw"};
```
mposa: Why ignore angular terms?

systems/controllers/affine_controller.cc
```diff
@@ -7,38 +7,40 @@ using Eigen::VectorXd;
...
+                                                   input_port_info_index_);
 
   const AffineParams* params = dynamic_cast<const AffineParams*>(
       this->EvalVectorInput(context, input_port_params_index_));
 
-  // could use MatrixXd instead of auto, but I think that would force a copy
-  auto K = params->get_K();
-  auto desired_state = params->get_desired_state();
-  auto E = params->get_E();
+  MatrixXd K = params->get_K();
```
mposa: Using `MatrixXd` here, I believe, forces a copy. `params` returns a `const Eigen::Map<Eigen::MatrixXd>`, which is a reference to the data (rather than a copy).

systems/controllers/constrained_lqr_controller.h
```diff
@@ -0,0 +1,121 @@
...
+#include "systems/framework/output_vector.h"
+
+namespace dairlib {
+namespace systems {
+
+/*
+ * ConstrainedLQRController class that implements an LQR controller that also
+ * takes into account constraints in the state space.
+ * The controller follows an affine structure. It is not a subclass of
+ * AffineControlelr due to difficulties in incorporating the required input
```
mposa: nit-typo

systems/controllers/constrained_lqr_controller.h
```diff
@@ -0,0 +1,121 @@
...
+ * is taken into account while computing the Jacobians. If it is not provided,
+ * the controller assumes no contact.
+ */
+class ConstrainedLQRController : public drake::systems::LeafSystem<double> {
+ public:
+  /*
+   * The constructor computes K and E and is the major computational block of
+   * the controller class.
+   */
+  ConstrainedLQRController(const RigidBodyTree<double>& tree, Eigen::VectorXd q,
```
mposa: All of these inputs should probably be const references.

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -0,0 +1,141 @@
...
+using drake::systems::Context;
+using drake::systems::controllers::LinearQuadraticRegulator;
+using drake::systems::controllers::LinearQuadraticRegulatorResult;
+
+ConstrainedLQRController::ConstrainedLQRController(
+    const RigidBodyTree<double>& tree, VectorXd q, VectorXd u, VectorXd lambda,
+    MatrixXd Q, MatrixXd R, ContactInfo contact_info)
+    : tree_(tree),
+      contact_info_(contact_info),
+      contact_toolkit_(
```
mposa: Question: why a unique_ptr?

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -0,0 +1,141 @@
...
+      num_positions_(tree.get_num_positions()),
+      num_velocities_(tree.get_num_velocities()),
+      num_states_(num_positions_ + num_velocities_),
+      num_efforts_(tree.get_num_actuators()),
+      num_forces_(tree.getNumPositionConstraints() +
+                  contact_info.num_contacts * 3) {
+  // Input port that takes in an OutputVector containing the current Cassie
+  // state
+  input_port_info_index_ =
+      this
```
mposa: Linebreak formatting looks off here

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -0,0 +1,141 @@
...
+          ->DeclareVectorInputPort(OutputVector<double>(
+              num_positions_, num_velocities_, num_efforts_))
+          .get_index();
+
+  // Output port that outputs the efforts
+  output_port_efforts_index_ =
+      this->DeclareVectorOutputPort(TimestampedVector<double>(num_efforts_),
```
mposa: Is this better than 
```
output_port_efforts_index_ =
    this->DeclareVectorOutputPort(TimestampedVector<double>(num_efforts_),
        &ConstrainedLQRController::CalcControl).get_index();
```
?

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -0,0 +1,141 @@
...
+  // Contact Jacobian (If contact information is provided)
+  if (contact_info_.num_contacts > 0) {
+    J_contact = autoDiffToValueMatrix(
+        contact_toolkit_->CalcContactJacobian(initializeAutoDiff(x)));
+  }
+
+  MatrixXd J(J_tree.rows() + J_contact.rows(), num_positions_);
+  J << J_tree, J_contact;
+
+  // Computing F
```
mposa: Add brief explanation of what `F` is

systems/controllers/constrained_lqr_controller.h
```diff
@@ -0,0 +1,121 @@
...
+#include "attic/multibody/contact_toolkit.h"
+#include "drake/systems/controllers/linear_quadratic_regulator.h"
+#include "drake/systems/framework/leaf_system.h"
+#include "systems/framework/output_vector.h"
+
+namespace dairlib {
+namespace systems {
+
+/*
+ * ConstrainedLQRController class that implements an LQR controller that also
```
mposa: Somewhere in this description would be a good place to reference the appropriate publication.

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -0,0 +1,141 @@
...
+  B_ = P * B;
+  // Remapping the Q costs to the new coordinates
+  Q_ = P * Q * P.transpose();
+  R_ = R;
+
+  // Validating the dimesions of A_, B_, Q_ and R_ with respect to each other.
+  DRAKE_DEMAND(Q_.rows() == Q_.cols());
+  DRAKE_DEMAND(R_.rows() == R_.cols());
+  DRAKE_DEMAND(A_.rows() == Q_.rows());
+  DRAKE_DEMAND(B_.cols() == R_.rows());
```
mposa: Aren't all of these checks guaranteed by the linear algebra above? I think the only one that isn't guaranteed is the last `B.cols == R.rows`. It's probably better to check that `Q` and `R` inputs have appropriate size, then assume that the matrix multiplies result in proper dimensions.

systems/controllers/constrained_lqr_controller.cc
shrenikm: I need `tree` and `contact_info` to create the object (It doesn't have a default constructor). So I couldn't really define it as a private non-pointer variable.

systems/controllers/constrained_lqr_controller.cc
shrenikm: For some reason my linting formats it this way. I'm using the Cland formatter.

mposa:
Is this waiting for a re-review? It does need to resolve conflicts with master.

shrenikm:
The quaternion base fixed point solver still needs to be fixed. I'll request a review once it's done and all conflicts have been resolved.

shrenikm:
## Updates

* Added quaternion floating base tests to `multibody_solvers_test`, `contact_toolkit_test` and `constrainted_lqr_controller_test`.
* Fixed Eigen bugs with the quaternion implementation of the solvers and controller.
* Made the changes requested in the previous review.
* Moved the `StateConnector` (Utility connector system that produces an `OutputVector` output given a `BasicVector` input of the state) to `multibody/utility_systems`.
* Updated the controller computations to fix bugs caused due to the quaternion models. Two separate Jacobians are computed now.
* `ContactToolkit` now has a `in_terms_of_qdot` boolean member that is used to compute the type of Jacobian. By default, if the number of positions and velocities are equal (Ex: RPY base), `in_terms_of_qdot=true`, else (Ex: Quaternion base) `in_terms_of_qdot=false`.
* `CalcContactJacobian` in `ContactToolkit` now is an overloaded function that can take in a `in_terms_of_qdot` boolean argument and compute the Jacobian accordingly. This is useful in the case where the `ContactToolkit` object is created using a specific tree, and both types of Jacobians are required to be computed from it. (See `systems/controllers/constrained_lqr_controller`).
* The controller simulation has been tested (Slow with contact).
* The added functionality and tests have been documented.

systems/controllers/affine_controller.cc
```diff
@@ -7,35 +7,38 @@ using Eigen::VectorXd;
...
+                                   TimestampedVector<double>* control) const {
+  const OutputVector<double>* info =
+      (OutputVector<double>*)this->EvalVectorInput(context,
+                                                   input_port_info_index_);
 
   const AffineParams* params = dynamic_cast<const AffineParams*>(
       this->EvalVectorInput(context, input_port_params_index_));
 
-  // could use MatrixXd instead of auto, but I think that would force a copy
+  // Using the Eigen type instead of auto would force a copy.
```
mposa: `auto` is still generating an Eigen type, but it's a Eigen `Map`, which acts as a reference. I think the original comment was more accurate.

systems/controllers/affine_controller.h
```diff
@@ -58,45 +61,49 @@ class AffineController : public drake::systems::LeafSystem<double> {
...
-      K_(get_mutable_data().head(num_states * num_efforts).data(), num_efforts,
-         num_states),
-      E_(get_mutable_data().segment(num_states * num_efforts, num_efforts)
-                                      .data(), num_efforts),
-      desired_state_(get_mutable_data().segment(num_states * num_efforts +
-                                                num_efforts, num_states).data(),
-                     num_states) {}
-
-  int GetNumStates() const { return num_states_;}
+  AffineParams(int num_states, int num_efforts)
```
mposa: Is this spacing really better?

attic/multibody/multibody_solvers.cc
```diff
@@ -614,7 +646,7 @@ void FixedPointSolver::AddJointLimitConstraint(const double tolerance) {
   VectorXd joint_min = tree_.joint_limit_min;
   VectorXd joint_max = tree_.joint_limit_max;
 
-  for (int i = 0; i < joint_min.size(); ++i) {
+  for (int i = 7; i < joint_min.size(); ++i) {
```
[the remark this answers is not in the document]
shrenikm: I've updated it to ignore the positions with names starting with "base". It isn't the most effective way of doing it but is probably better than the hardcoded version.

systems/controllers/affine_controller.h
shrenikm: This (And the other formatting differences) is the output of running my clang formatter on the file. Not really sure what to do about it.

PR #61, opened 2019-04-17
Add back Cassie RBT visualizer and fix a simulation bug

yminchen:
- Added back `cassie_state_visualizer.cc` for floating base RBT cassie
- Fixed a bug in `run_simple_sim.cc`


mposa:
Going to close this, as it seems superseded by #62.

PR #62, opened 2019-04-19
Update dispatcher

yminchen:
Building upon #61. 
Added `floating_base` flag so that `dispatcher_robot_out` and `dispatcher_robot_in` can take either fixed-base or floating-base Cassie.


examples/Cassie/run_simple_sim.cc
```diff
@@ -165,6 +165,10 @@ int do_main(int argc, char* argv[]) {
   Eigen::VectorXd x0 =
       Eigen::VectorXd::Zero(plant->get_rigid_body_tree().get_num_positions() +
                             plant->get_rigid_body_tree().get_num_velocities());
+  if(FLAGS_floating_base)
+    if(x0(4)==0 && x0(5)==0 && x0(6)==0)
```
mposa: Why does this check x0(4), x0(5), x0(6)? Aren't these angular velocity? Not sure why they're relevant. Also, the line before sets x0=0.

yminchen: `[x(3), x(4), x(5), x(6)]` is the quaternion

mposa: Ah, of course. I don't think this is the appropriate check then. It's meaningless as-is now, because we know `x0=0`. However, if something changed down the line and the `if` statement was false, we still might end up with an invalid quaternion. A more fool-proof change would be

`x0(3) = sqrt(1-x0(4)^2-x0(5)^2-...)`

yminchen: What if the user enters a non-unit quaternion?

yminchen: The purpose of this change is just to prevent people from carelessly initialize a 0-norm quaternion which doesn't represent any rotation and causes runtime error in Drake.

PR #70, opened 2019-06-13
Update dairlib master branch

yminchen:
Update dairlib so it keeps up with latest drake

- changed `CalcBiasForFrameGeometricJacobianExpressedInWorld` to  `CalcBiasForJacobianSpatialVelocity`
- calling a different function that creates gravity in MBP
- removed `ReconstructInputTrajectory()` and `ReconstructStateTrajectory()`


yminchen:
> Looks good to me. Thank you.

No problem!

Issue #63, opened 2019-04-24
The output of Cassie state estimator is stale


yminchen:
In `cassie-state-estimator` branch. Commit # `1fc510af1c9625aa555a2a9698089c7d6da69f1f`

The output of `CassieRbtStateEstimator` (a child of `LeafSystem`) is not computed from the latest Lcm message, but from the previous one. 

It seems that the per-step update `CassieRbtStateEstimator::Update()` is not called before the output function `CassieRbtStateEstimator::CopyStateOut()`. 
Experiment: We stored `context.time` into a discrete state of `CassieRbtStateEstimator` in `Update()`, and printed the following two things in `CopyStateOut()` 
- `context.time` and 
- the time stored as a discrete state.

Result: `context.time` is always bigger than the time stored as a discrete state.

Procedures to reproduce the experiment result:
- Build `examples/Cassie:dispatcher_robot_out` and `examples/Cassie:run_simple_sim`.
- In procman, run `2.a.dispatcher-robot-out (drake)` and then `1.simulator (dispatched)`.
- Look at the message printed by `dispatcher_robot_out`.

yminchen:
It turned out that `context.time` lags behind `lcm.time`, so we cannot use `context.time` to determine whether we should run the update step of EKF or not.

yminchen:
One way to fix this is to add an extra output port that only passes `lcm.time`, so `StateEstimator` has access to `lcm.time`. 

However, we will ignore the issue for now, since TRI people is modifying `LcmDrivenLoop`.

yminchen:
Found the reason why context.time lags behind lcm.time. 

Drake documentation (Class `Simulator` and Section **How the simulation is stepped**) shows the pseudocode that does the discrete update. The code is
    `xd⁺(tₛ) ← DoAnyDiscreteUpdates(tₛ, x*(tₛ))`
where `tₛ` is the start time of the step.  
Since they feed `tₛ` into the update function, we don't get the end time from `context.time`.

mposa:
Per conversation with Evan and Russ on slack, two possible approaches are suggested.

1. When a new message is received, and before the simulator is advanced, register a new `kTimed` event at `t=lcm.time`. The event handler would be the EKF update.
2. Alternatively, declare the EKF update event as a `ForcedEvent` (using `DeclareForcedDiscreteUpdateEvent`). After advancing the simulation, call `CalcDiscreteVariableUpdates`.

[...]

PR #74, opened 2019-06-27
Cassie state estimator

yminchen:
**What's new:**
1. Modified the structure so it works with both fixed-base and floating-base models. 

There are two main functions, `CopyStateOut()` and `Update()`, in `cassie_rbt_state_estimator`. We only declare the update event when the model is floating-based. Therefore, `cassie_rbt_state_estimator` acts as a passthrough if the model is fixed-based. 

2. Added a function which solves for the unknown joints (heel spring joint). Also added a unit test function for this.

3. Added contact estimation in `cassie_rbt_state_estimator`.

We say a foot is on the ground if the spring deflection is over a threshold. If it's under the threshold, we use 3 quadratic programs (with different contacts assumptions) to improve the estimation. If all 3 costs are high, we believe it's going through impact event, and we assume it's double support. The current contact estimator assumes that
- the swing leg doesn't stop during single support
- the robot doesn't transition from flight phase to single support phase

4. Added actuation feedback to `OutputVector`.

5. [minor update] Added terrain in floating-base simulation, and assign initial robot height if it's not set by users.

6. [minor update] Added a `floating-base` flag for `dispatcher_robot_out.cc`.

7. [minor update] Updated run_constrained_lqr_controller_sim.cc, so that it can run with the dispatcher.

**Ongoing:**
1. Unit test for contact estimation
2. Merge master's latest commit after it compiled successfully

**Next steps:**
1. Fix the issue of one time step delay in `Update()`
2. EKF debugging
3. Delete the codes used for testing during development. (e.g. cassie_state receiver)


examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -10,24 +10,111 @@
 #include "systems/framework/output_vector.h"
 #include "systems/framework/timestamped_vector.h"
 #include "examples/Cassie/datatypes/cassie_out_t.h"
+#include "examples/Cassie/cassie_utils.h"
 
 namespace dairlib {
 namespace systems {
 
+using Eigen::Vector3d;
```
mposa: Don't use `using` in a header. Everyone who `#incude`'s this header, would then be stuck with your `using` definitions.

examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -10,24 +10,111 @@
...
+  explicit CassieRbtStateEstimator(const RigidBodyTree<double>&,
+                                   bool is_floating_base);
+  void solveFourbarLinkage(double* left_heel_spring,
+                           double* right_heel_spring,
+                           const VectorXd& q_init) const;
+
  private:
-  void Output(const drake::systems::Context<double>& context,
-      OutputVector<double>* output) const;
+  void AssignImuValueToOutputVector(
```
mposa: Here and elsewhere. Having the output pointer be the first argument is a little awkward. Would prefer it be the second/last argument.

examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -10,24 +10,111 @@
...
-  std::map<std::string, int> velocityIndexMap_;
-  std::map<std::string, int> actuatorIndexMap_;
+  const bool is_floating_base_;
+
+  std::map<std::string, int> position_index_map_;
+  std::map<std::string, int> velocity_index_map_;
+  std::map<std::string, int> actuator_index_map_;
+
+  // Body indices
+  int left_thigh_ind_ = -1;
```
mposa: Why assigned to -1?

examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -10,24 +10,111 @@
...
+  DiscreteStateIndex state_idx_;
+  DiscreteStateIndex ekf_X_idx_;
+
+  DiscreteStateIndex previous_velocity_idx_;
+
+  DiscreteStateIndex filtered_residue_double_idx_;
+  DiscreteStateIndex filtered_residue_left_idx_;
+  DiscreteStateIndex filtered_residue_right_idx_;
+
+  DiscreteStateIndex ddq_double_init_idx_;
```
mposa: Many of these variables are not totally clear by their name. Some comments would be helpful.

examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -10,24 +10,111 @@
...
+  Vector3d gravity_ = Vector3d(0, 0, -9.81);
+
+  // Contact Estimation Parameters
+  const double cost_threshold_ = 200;
+  const double knee_spring_threshold_ = -0.015;
+  const double heel_spring_threshold_ = -0.03;
+  const double eps_cost_ = 1e-10;  // Avoid indefinite matrix
+  const double eps_imu_ = 0.5;  // Error bounds for imu acceleration
+  const double w_soft_constraint_ = 100;  // soft constraint cost
+  const double alpha_ = 0.9;  // Decay for residue calculation
```
mposa: residual?

examples/Cassie/networking/cassie_output_receiver.cc
```diff
@@ -123,6 +123,9 @@ void CassieOutputReceiver::CopyOutput(const Context<double>& context,
   copy_vector(message.messages,
               cassie_out->messages, 4);
   cassie_out->isCalibrated = message.isCalibrated;
+
+  // Just for testing
```
mposa: Delete dead commented code

examples/Cassie/run_simple_sim.cc
```diff
@@ -228,6 +233,13 @@ int do_main(int argc, char* argv[]) {
 
   std::cout << q0 << std::endl;
 
+  // If it's floating base and the initial height is 0, we adjust it to be above
+  // the ground.
+  if (FLAGS_floating_base){
+    if (x0(2) == 0)
+      x0(2) = 1.5;
```
mposa: This is a strange sequence. Why `if x0(2) == 0`? What if it's close to zero? Why 1.5?

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+    lambda_b_left_init_idx_ = DeclareDiscreteState(VectorXd::Zero(2, 1));
+    lambda_b_right_init_idx_ = DeclareDiscreteState(VectorXd::Zero(2, 1));
+    lambda_cl_double_init_idx_ = DeclareDiscreteState(VectorXd::Zero(6, 1));
+    lambda_cl_left_init_idx_ = DeclareDiscreteState(VectorXd::Zero(6, 1));
+    lambda_cr_double_init_idx_ = DeclareDiscreteState(VectorXd::Zero(6, 1));
+    lambda_cr_right_init_idx_ = DeclareDiscreteState(VectorXd::Zero(6, 1));
+  }
+}
+
+void CassieRbtStateEstimator::solveFourbarLinkage(
```
mposa: This function is going to be hard for anyone to understand/debug. It could use a *lengthy* block of documentation describing how it works.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+  output->SetEffortAtIndex(actuator_index_map_.at("knee_right_motor"),
+                           cassie_out.rightLeg.kneeDrive.torque);
+  output->SetEffortAtIndex(actuator_index_map_.at("toe_right_motor"),
+                           cassie_out.rightLeg.footDrive.torque);
+}
+
+void CassieRbtStateEstimator::AssignNonFloatingBaseStateToOutputVector(
+    OutputVector<double>* output, const cassie_out_t& cassie_out) const {
+  // Copy the robot state excluding floating base
+  // TODO(yuming): check what cassie_out.leftLeg.footJoint.position is.
```
mposa: Is this TODO complete or still TODO? Seems important.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+  output->SetVelocityAtIndex(velocity_index_map_.at("base_vx"), est_fb_state(10));
+  output->SetVelocityAtIndex(velocity_index_map_.at("base_vy"), est_fb_state(11));
+  output->SetVelocityAtIndex(velocity_index_map_.at("base_vz"), est_fb_state(12));
+}
+
+
+EventStatus CassieRbtStateEstimator::Update(const Context<double>& context,
+    DiscreteValues<double>* discrete_state) const {
+  // Testing
+  /*const auto& cassie_out =
```
mposa: Remove dead code

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+  Isometry3d pelvis_pose = tree_.CalcBodyPoseInWorldFrame(cache,
+      *pelvis_body);
+  MatrixXd R_WB = pelvis_pose.linear();
+  Vector3d gravity_in_pelvis_frame = R_WB.transpose()*gravity_;
+  alpha_imu -= gravity_in_pelvis_frame;
+
+  std::vector<double> optimal_cost;
+
+  // Mathematical program - double contact
+  MathematicalProgram quadprog_double;
```
mposa: This QP should be created once, not inside the loop. In the loop, just update it and solve. This is probably something we can leave, provided someone creates an issue to resolve it soon.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+        lambda_b_double_init_idx_).get_mutable_value() <<
+        result_double.GetSolution(lambda_b);
+    discrete_state->get_mutable_vector(
+        lambda_cl_double_init_idx_).get_mutable_value() <<
+        left_force;
+    discrete_state->get_mutable_vector(
+        lambda_cr_double_init_idx_).get_mutable_value() <<
+        right_force;
+
+    // Residue calculation
```
mposa: Residual (here and in variable naming)

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+      current_time;
+  }
+  return EventStatus::Succeeded();
+}
+
+
+// Contact Estimator assumes:
+// 1. the swing leg doesn't stop during single support
+// 2. the robot doesn't transition from flight to single support
+void CassieRbtStateEstimator::contactEstimation(
```
mposa: Needs documentation for what the algorithm is doing.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+    *left_contact = 1;
+    *right_contact = 1;
+  } else if(min_index == 1) {
+    *left_contact = 1;
+  } else if(min_index == 2) {
+    *right_contact = 1;
+  }
+
+  // Update contact estimation based on spring deflection information
+  // We say a foot is in contact with the ground if sprig deflection is over
```
mposa: sprig->spring

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+/// estimated state as an OutputVector
 /// Since it needs to map from a struct to a vector, and no assumptions on the
 /// ordering of the vector are made, utilizies index maps to make this mapping.
-void CassieRbtStateEstimator::Output(
-    const Context<double>& context, OutputVector<double>* output) const {
+void CassieRbtStateEstimator::CopyStateOut(
+  const Context<double>& context, OutputVector<double>* output) const {
   const auto& cassie_out =
-      this->EvalAbstractInput(context, 0)->get_value<cassie_out_t>();
+    this->EvalAbstractInput(context, 0)->get_value<cassie_out_t>();
```
mposa: This spacing looks off.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
 
 /// Workhorse state estimation function. Given a `cassie_out_t`, compute the
-/// esitmated state as an OutputVector
+/// estimated state as an OutputVector
 /// Since it needs to map from a struct to a vector, and no assumptions on the
 /// ordering of the vector are made, utilizies index maps to make this mapping.
-void CassieRbtStateEstimator::Output(
-    const Context<double>& context, OutputVector<double>* output) const {
+void CassieRbtStateEstimator::CopyStateOut(
+  const Context<double>& context, OutputVector<double>* output) const {
```
mposa: New spacing looks off.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,140 +1,981 @@
...
+                              cassie_state->GetVelocities()[3]);
+    output->SetVelocityAtIndex(velocity_index_map_.at("base_vy"),
+                              cassie_state->GetVelocities()[4]);
+    output->SetVelocityAtIndex(velocity_index_map_.at("base_vz"),
+                              cassie_state->GetVelocities()[5]);
+  }
+
+
+
+  // Testing
```
mposa: Delete dead code.

examples/Cassie/cassie_rbt_state_estimator.h
yminchen: Done!

examples/Cassie/cassie_rbt_state_estimator.h
yminchen: -1 was the value that indicates there is no such body name. I added it because I didn't know that the function `GetBodyIndexFromName` already returns -1 if the body name doesn't exist. I'll delete the initial value.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: This is still a TODO. We cannot check it in simulation, because we create `cassie_out` ourselves. I'm wondering what `cassie_out.leftLeg.footJoint.position` in Cassie hardware.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Created an issue that aims to resolve this. #76

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Is it ok that I leave the codes here because I'm going to use it to fix the "one time-step delay" problem #63.
I added `TODO(yminchen)` to all the codes that should be removed at the end, so I can delete them in the future.

examples/Cassie/run_simple_sim.cc
yminchen: The idea is to prevent someone in the future who wants to use `run_simple_sim.cc` with floating-base model but forgets to set the initial height of the robot

yminchen: 1.5 is just a random selected number that guarantees the robot is above the ground, so when the simulation runs the robot is not below the ground

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Done!

yminchen:
@mposa I've made changes with a few questions in the conversations.

examples/Cassie/run_simple_sim.cc
mposa: I'm not sure if I buy the reasoning for this. The user needs to set the initial configuration, somehow. I could see something like
```
if (FLAGS_floating_base)
 x0(2) = 1.5
```
being used right after `x0` is first created.

yminchen: Yes, I can change to that

mposa:
Looks good to me. Is it ready to merge?

yminchen:
> Looks good to me. Is it ready to merge?

Yes, it's ready

PR #79, opened 2019-07-01
Generic lcm log parser

Nanda-Kishore-V:
**What's new:**
- Implement a generic lcm log parser function to convert lcm log into matrices made up of the parsed vectors. The function is templated on the lcm type of the message and the class to convert the lcm message to a vector.


attic/multibody/generic_lcm_log_parser.h
```diff
@@ -0,0 +1,52 @@
...
+    drake::systems::Simulator<double> sim(*diagram);
+
+    sim.StepTo(r_log.GetNextMessageTime() + duration);
+
+    *t = output_aggregator->BuildTimestampVector();
+
+    *x = output_aggregator->BuildMatrixFromVectors();
+  }
+} // multibody
+} //dairlib
```
mposa: Missing carriage return at end of the file.

attic/multibody/generic_lcm_log_parser.h
```diff
@@ -0,0 +1,52 @@
...
+    drake::lcm::DrakeLcmLog r_log(file, false);
+
+    drake::systems::DiagramBuilder<double> builder;
+
+    builder.AddSystem<drake::systems::lcm::LcmLogPlaybackSystem>(&r_log);
+
+    auto output_sub = builder.AddSystem(
+        LcmSubscriberSystem::Make<T>(channel, &r_log));
+
+    auto output_receiver = builder.AddSystem<U>(tree);
```
mposa: This requires that `U(tree)` be a constructor. Is it possible to be more general? For instance, what if, the argument was `parseLcmLog(std::unique_ptr<U> system`, then you should be able to call `builder.AddSystem(system)` using this API

https://github.com/RobotLocomotion/drake/blob/master/systems/framework/diagram_builder.h#L56

This would also allow this to apply to any system with vector outputs (and should therefore be moved to `systems/`

Nanda-Kishore-V: Do you think the implementation can be extended to `builder.AddSystem(Args &&... args)` as well?

mposa: It probably could be, but would that be a useful API?

Nanda-Kishore-V: Should these files be in `attic/systems/` or `systems/`?

mposa:
Looks great, thank

PR #81, opened 2019-07-12
Unit test for contact estimation

Nanda-Kishore-V:
Add unit tests for `contactEstimation`.


mposa:
Looks good to me--I had one suggestion but I wanted to make sure it would work first. I ended up pushing the changes. You can use `system.CreateDefaultContext()` to automatically build a context, and then extract what you want from that context. In this case, I used `context.get_mutable_discrete_state()`

PR #85, opened 2019-07-21
ROS Publishers and Subscribers

mposa:
~~This is a work in progress, and is not yet meant to be merged.~~

This PR is a port of two other projects:

1. https://github.com/nicolov/ros-bazel
2. https://github.com/gizatt/drake_ros_systems

**ros-bazel** builds ROS within bazel using catkin. Currently, this requires running an installation script.
**drake_ros_systems** is an implementation of ROS pub/sub systems.

Neither of these libraries are currently in use, and both required updating to work with the latest versions of bazel and drake.

As a goal, dairlib should not *require* ROS. The library should build and run properly, though ROS-specific libraries will not function without an installation.

## Tasks for this PR
- [x]  Port of **ros-bazel**
- [x]  Port of **drake_ros_systems**
- [x] Create unified installation documentation
- [x] Gracefully skip ROS code if it is not enabled
- [x] CI integration both with and without ROS

## Installation
1. install ROS, and don't forget add to `.bashrc`
```
export ROS_MASTER_URI=http://localhost:11311
source /opt/ros/melodic/setup.bash 
```
2. Install some dependencies
```
sudo apt install python-rosinstall-generator python-catkin-tools
```
2. In `dairlib/tools/workspace/ros`, build the ros workspace using catkin
```
./compile_ros_workspace.sh
```

## Notes
- Environments that require ROS should be marked with the tag `ros`, e.g.
```
cc_library(
...
  tags=["ros"],
)
```
-The environment variable `DAIRLIB_WTH_ROS` must be set to `ON`. If it is not, then all environments with the tag `ros` will be skipped, by default, during a build or test. This can be done by adding to `~/.bashrc`
```
export DAIRLIB_WITH_ROS=ON
```

### DIFF
--- a/systems/ros/ros_publisher_system.h
+++ b/systems/ros/ros_publisher_system.h
@@ -101,7 +101,10 @@ class RosPublisherSystem : public drake::systems::LeafSystem<double> {
         (trigger == TriggerType::kPerStep));
     }
 
+    // Outgoing queue size chosen to be small, but 5 is arbitrary
     publisher_ = node_handle->advertise<RosMessage>(topic, 5);
+    // check that publisher is not empty
+    DRAKE_THROW_UNLESS(publisher_);
 
     DeclareAbstractInputPort("ros_message", drake::Value<RosMessage>());
     set_name(make_name(topic_));
--- a/systems/ros/ros_publisher_system.h
+++ b/systems/ros/ros_publisher_system.h
@@ -1,0 +1,109 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "drake/common/drake_copyable.h"
+#include "drake/lcm/drake_lcm_interface.h"
+#include "drake/systems/framework/leaf_system.h"
+
+#include "ros/ros.h"
+
+namespace drake_ros_systems {
+
+using namespace drake;
+
+/**
+ * Publishes an ROS message containing information from its input port.
+ *
+ * @ingroup message_passing
+ */
+template <typename RosMessage>
+class RosPublisherSystem : public systems::LeafSystem<double> {
+ public:
+  DRAKE_NO_COPY_NO_MOVE_NO_ASSIGN(RosPublisherSystem)
+
+  /**
+   * A factory method that returns an %RosPublisherSystem that takes
+   * Value<RosMessage> message objects on its sole abstract-valued input port.
+   *
+   * @tparam RosMessage message type to serialize.
+   *
+   * @param[in] topic The ROS topic on which to publish.
+   *
+   * @param node_handle The ROS context.
+   */
+  static std::unique_ptr<RosPublisherSystem<RosMessage>> Make(
+      const std::string& topic, ros::NodeHandle* node_handle) {
+    return std::make_unique<RosPublisherSystem<RosMessage>>(topic, node_handle);
+  }
+
+  // TODO(gizatt): add multiple DrakeRosInterface, so you can publish to a log
+  // and real lcm at the same time. (TODO cloned from LCM publisher system,
+  // originally attributed to Siyuan.)
+
+  /**
+   * A constructor for an %RosPublisherSystem that takes ROS message objects on
+   * its sole abstract-valued input port. The ROS message type is determined by
+   * the template type.
+   *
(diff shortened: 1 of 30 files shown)


systems/ros/ros_publisher_system.h
```diff
@@ -0,0 +1,198 @@
...
+    using drake::systems::TriggerType;
+
+    // Check that publish_triggers does not contain an unsupported trigger.
+    for (const auto& trigger : publish_triggers) {
+      DRAKE_THROW_UNLESS((trigger == TriggerType::kForced) ||
+        (trigger == TriggerType::kPeriodic) ||
+        (trigger == TriggerType::kPerStep));
+    }
+
+    publisher_ = node_handle->advertise<RosMessage>(topic, 5);
```
mshalm: Is it overkill here to double check that publisher_ isn't empty? the [ROS wiki page](http://wiki.ros.org/roscpp/Overview/Publishers%20and%20Subscribers) recommends checking.

Is 5 the right queue length? is that a standard value for implementing a publisher with ROS?

systems/ros/ros_publisher_system.h
```diff
@@ -0,0 +1,198 @@
...
+   *
+   * @param publish_triggers Set of triggers that determine when messages will
+   * be published. Supported TriggerTypes are {kForced, kPeriodic, kPerStep}.
+   * Will throw an error if empty or if unsupported types are provided.
+   *
+   * @param publish_period Period that messages will be published (optional).
+   * publish_period should only be non-zero if one of the publish_triggers is
+   * kPeriodic.
+   */
+  RosPublisherSystem(const std::string& topic, ros::NodeHandle* node_handle,
```
mshalm: Any particular reason for implementing a large function in the header file?

systems/ros/ros_subscriber_system.h
```diff
@@ -0,0 +1,235 @@
...
+   * @param[in] channel The ROS topic on which to subscribe.
+   *
+   * @param node_handle The ROS context.
+   */
+  RosSubscriberSystem(const std::string& topic, ros::NodeHandle* node_handle)
+      : topic_(topic), node_handle_(node_handle) {
+    DRAKE_DEMAND(node_handle_);
+
+    subscriber_ = node_handle->subscribe(
+        topic, 100, &RosSubscriberSystem<RosMessage>::HandleMessage, this);
```
mshalm: Similar question to publisher system on static queue size

systems/ros/ros_subscriber_system.h
```diff
@@ -0,0 +1,235 @@
...
+      drake::systems::AbstractValues* abstract_state) const {
+    std::lock_guard<std::mutex> lock(received_message_mutex_);
+    abstract_state->get_mutable_value(kStateIndexMessage)
+        .get_mutable_value<RosMessage>() = received_message_;
+    abstract_state->get_mutable_value(kStateIndexMessageCount)
+        .get_mutable_value<int>() = received_message_count_;
+  };
+
+  // Callback entry point from ROS into this class. Also wakes up one thread
+  // block on notification_ if it's not nullptr.
```
mshalm: Not clear what this condition on "not being nullptr" is as there's no apparent conditional logic in the function implementation.

systems/ros/ros_publisher_system.h
mposa: It's implemented in a header because it's templated on message type, but `ros_publisher_system` does not know, at compile time, what explicit types to compile (only those that link against it do).

systems/ros/ros_publisher_system.h
mposa: -Added a check.
-5 is pretty arbitrary. I added a comment mentioning this. The queue should be small, but there's nothing special about 5.

systems/ros/ros_subscriber_system.h
mposa: This queue should probably be of size 1, as it is with LCM.

systems/ros/ros_subscriber_system.h
mposa: Not sure what that comment means either. Deleted it.

PR #87, opened 2019-07-22
Added rigidbodytree accelerometer

yminchen:
Drake's accelerometer has a bug -- the sign of gravitational force contributed to accelerometer measurement is incorrect. The measured acceleration caused by gravity should point to positive z direction.
https://github.com/RobotLocomotion/drake/blob/d72ad8f988f9c0d002959d00e1b7002daedaff41/attic/systems/sensors/accelerometer.cc#L133

Since it's a quick fix and RBT will be deprecated soon, I just added it to dairlib. We will be able to use it to test Cassie's controller in simulation. 

Changes:
- Added RbtAccelerometer
- Updated the frame of Cassie's IMU.

### DIFF
--- a/examples/Cassie/BUILD.bazel
+++ b/examples/Cassie/BUILD.bazel
@@ -34,6 +34,7 @@ cc_library(
         "//systems/primitives",
         "//systems:robot_lcm_systems",
         "//systems/sensors:sim_cassie_sensor_aggregator",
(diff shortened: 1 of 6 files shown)


mposa:
Going to close this, as it's transitioned to https://github.com/RobotLocomotion/drake/pull/11821

Feel free to re-open if I'm closing it by error.

PR #90, opened 2019-07-28
Cassie sim analysis

zhifeishen99:
tidied up scripts


mposa:
You've got a ton of conflicts here, I'm guessing mostly with files that you've never changed, due to a mistake somewhere in the merge process.

mposa:
Just took a quick look to try to sort out the conflicts, and the code isn't really clean enough for me to do so. The fact that GitHub here is saying there are 496 files changed is a sign that things haven't gone smoothly! 

Some initial steps required:

1. Merge back to master, and resolve files with conflicts. Mostly, this means taking whatever's in master.
2. Remove duplicate files (I see many instances here of the same files)
3. Remove the data files. These should never be checked into the repository.

PR #94, opened 2019-08-09
Osc testing and debugging

mposa:
Minor changes here--and (what I think are) some bug fixes to the OSC.

### DIFF
--- a/attic/multibody/multibody_solvers.cc
+++ b/attic/multibody/multibody_solvers.cc
@@ -250,8 +250,12 @@ void PositionSolver::AddJointLimitConstraint(const double tolerance) {
 
   for (int i = 0; i < joint_min.size(); ++i) {
     // Adding minimum and maximum joint angle constraints.
-    prog_->AddConstraint(q_(i) >= (joint_min(i) + tolerance));
-    prog_->AddConstraint(q_(i) <= (joint_max(i) - tolerance));
+    // Ignoring the floating base coordinates if any.
+    // We assume all floating base positions start with "base"
+    if (tree_.get_position_name(i).substr(0, 4) != "base") {
+      prog_->AddBoundingBoxConstraint(joint_min(i) + tolerance,
+        joint_max(i) - tolerance, q_(i));
+    }
   }
 }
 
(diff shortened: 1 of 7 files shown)


attic/multibody/multibody_solvers.cc
```diff
@@ -648,8 +651,8 @@ void FixedPointSolver::AddJointLimitConstraint(const double tolerance) {
     // Ignoring the floating base coordinates if any.
     // We assume all floating base positions start with "base"
     if (tree_.get_position_name(i).substr(0, 4) != "base") {
-      prog_->AddConstraint(q_(i) >= (joint_min(i) + tolerance));
-      prog_->AddConstraint(q_(i) <= (joint_max(i) - tolerance));
+      prog_->AddBoundingBoxConstraint(joint_min(i) + tolerance,
```
yminchen: Assuming `tolerance` > 0, shouldn't this be 
prog_->AddBoundingBoxConstraint(joint_min(i) - tolerance, joint_max(i) + tolerance, q_(i)); ?

systems/controllers/osc/operational_space_control.cc
```diff
@@ -260,17 +260,17 @@ void OperationalSpaceControl::Build() {
   // Add costs
   // 1. input cost
   if (W_input_.size() > 0) {
-    prog_->AddQuadraticCost(MatrixXd::Identity(n_u_, n_u_),
+    prog_->AddQuadraticCost(W_input_,
```
yminchen: Can't believe there is such a big bug, since the robot is balancing and walking in simulation!

attic/multibody/multibody_solvers.cc
mposa: `tolerance` is really a misnomer. this is more of a margin on the joint limit, so that the fixed point isn't right up against it. I'll rename.

yminchen: I see. Then probably the word `relaxation` here (https://github.com/DAIRLab/dairlib/blob/75c10731894aabb7e64d0bcb2e1f51694ef45fc6/attic/multibody/multibody_solvers.h#L173) can also be changed to `tightening` for example.

PR #95, opened 2019-08-11
Osc testing--do not merge

mposa:
This commit should not be merged, but does demonstrate balancing in Drake and MuJoCo. I'm sure that not all of the gain changes here are necessary. To be merged, this requires a better way to either (1) set the individual costs of slack variables or (2) to ignore redundant constraints.


yminchen:
This PR is superseded by #130

PR #96, opened 2019-08-13
Invariant ekf

Nanda-Kishore-V:
**What's new:**
* Add Invariant EKF to estimate the position and orientation of the pelvis
* Modify the logic of contact estimation and allow different thresholds to be set for control and estimation.


mposa:
Per a face to face discussion with Nanda last week, we weren't sure where the lightly-modified third-party EKF code should be located. The goals are to:

- Easily incorporate upstream changes.
- Clearly mark which code is derived from Ross' code, and does not need to align with our code style practices.
- Allow, if necessary, users to change the EKF code during testing.

The best option we came up with was to fork Ross Hartley's code, and make changes in that branch (which will make it easy to later merge upstream changes). Include this repository as a pegged external with optional local installation, like we do with Drake.

@yminchen @Nanda-Kishore-V @yangwill Thoughts on this? We wanted to sleep on it, to ensure we aren't adding too much complexity. Having too many of these external repositories might, ultimately, get confusing.

mposa:
@yminchen @Nanda-Kishore-V @yangwill Following up on this. It'd be good to get this into a form everyone's happy with, and merge it.

yminchen:
@mposa @Nanda-Kishore-V, @yangwill and I discussed. We think that it's a good idea to fork Ross's code, add our own changes, and use this repo as a pegged external.

Nanda-Kishore-V:
**What's new:**
`invariant-ekf` is moved to its own (forked) repository. The default behavior is to use the pegged version of this `invariant-ekf` repository. However, it can also be downloaded as a local repository for making changes.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -816,32 +848,164 @@ void CassieRbtStateEstimator::contactEstimation(
...
+///   2. three optimizations one for each double support, left support and
+///   right support
+///  If the compression in the left (right) heel/ankle spring is more than the
+///  set threshold, the left (right) foot is estimated to be in contact with
+///  the ground.
+///  During impact, both the legs have some non-zero acceleration and hence the
+///  optimal costs will all be high. This case is assumed to be *no* stance.
+void CassieRbtStateEstimator::EstimateContactForEkf(
+    const OutputVector<double>& output,
+    int* left_contact, int* right_contact) const {
```
mposa: Should `optimal_cost` instead be an input to this method? This function seems to require that optimal_cost be set prior to calling it, which is always a dangerous requirement (and currently undocumented)

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -816,32 +848,164 @@ void CassieRbtStateEstimator::contactEstimation(
...
+///    actuator torques of the robot
+/// Ouput: left contact `left_contact` and right contact `right_contact` that
+///  indicate if the corresponding foot is in contact with the ground
+///
+/// Assumptions:
+///  1. the swing leg doesn't stop during single support
+///  2. flight mode is not present during the gait
+///
+/// Algorithm:
+///  The contact is estimated based on
```
mposa: Same as above comment regarding the algorithm

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -850,147 +1014,318 @@ void CassieRbtStateEstimator::contactEstimation(
...
+        cout << "Ground Truth: " << endl;
+        cout << "Positions: " << endl;
+        cout << imu_pos_wrt_world.transpose() << endl;
+        cout << "Orientation (quaternion) : " << endl;
+        cout << output_gt.GetPositions().segment<4>(3).transpose() << endl;
+        cout << "Velocities: " << endl;
+        cout << imu_vel_wrt_world.transpose() << endl;
+      }
+
+      // ofile.open("/home/nanda/DAIR/plotting/ekf.csv",
```
mposa: This output file logging should not be committed to master

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -850,147 +1014,318 @@ void CassieRbtStateEstimator::contactEstimation(
...
-    // discrete_state->get_mutable_vector(time_idx_).get_mutable_value() = ...;
+    // Step 4 - EKF (measurement step)
+    KinematicsCache<double> cache = tree_.doKinematics(
+        filtered_output.GetPositions(), filtered_output.GetVelocities(), true);
+    Vector3d d, p;
+    // rotation part of pose and covariance is unused in EKF
+    Eigen::Matrix4d pose = Eigen::Matrix4d::Identity();
+    Eigen::Matrix<double, 6, 6> covariance = MatrixXd::Identity(6, 6);
+    Eigen::Matrix<double, 22, 22> cov_w =
+        0.000289 * MatrixXd::Identity(22, 22); // nosie of joints measurement
```
mposa: Should this constant be set with the rest of the constants, rather than inline here?

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -850,147 +1014,318 @@ void CassieRbtStateEstimator::contactEstimation(
...
+    if (test_with_ground_truth_state_) {
+      if (print_info_to_terminal_) {
+        cout << "z difference: " <<
+            filter.getState().getPosition()[2] - imu_pos_wrt_world[2] << endl;
+      }
 
-    // Question: Do we need to filter the gyro value?
-    // We will get the bias (parameter) from EKF
+      for (int i = 0; i < 3; ++i) {
+        ofile << filter.getState().getPosition()[i] << ", ";
```
mposa: Ditto with this output file logging.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,5 +1,11 @@
+// TODO(yminchen): do we need to filter the gyro value before we publish?
+// We can get the bias (parameter) from EKF
+
+// TODO(yminchen): ask nanda why heel_spring_threshold_ctrl_ is 0.03?
+
 #include "examples/Cassie/cassie_rbt_state_estimator.h"
 #include <math.h>
+#include <utility>
 #include <fstream>
```
yangwill: Don't need to include fstream anymore

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -47,33 +62,75 @@ CassieRbtStateEstimator::CassieRbtStateEstimator(
...
-  right_thigh_ind_ = GetBodyIndexFromName(tree, "thigh_right");
-  left_heel_spring_ind_ = GetBodyIndexFromName(tree, "heel_spring_left");
-  right_heel_spring_ind_ = GetBodyIndexFromName(tree, "heel_spring_right");
-  DRAKE_DEMAND(left_thigh_ind_ != -1 && right_thigh_ind_ != -1 &&
-               left_heel_spring_ind_ != -1 && right_heel_spring_ind_ != -1);
+  left_thigh_idx_ = GetBodyIndexFromName(tree, "thigh_left");
+  right_thigh_idx_ = GetBodyIndexFromName(tree, "thigh_right");
+  left_heel_spring_idx_ = GetBodyIndexFromName(tree, "heel_spring_left");
+  right_heel_spring_idx_ = GetBodyIndexFromName(tree, "heel_spring_right");
+  DRAKE_DEMAND(left_thigh_idx_ != -1 && right_thigh_idx_ != -1 &&
```
yangwill: Although it's good to check that the indices are valid, I feel like this check should be inside GetBodyIndexFromName

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -816,32 +848,164 @@ void CassieRbtStateEstimator::contactEstimation(
...
+    const OutputVector<double>& output,
+    int* left_contact, int* right_contact) const {
   // Estimate contact based on optimization results
   // The vector optimal_cost has double support, left support and right support
   // costs in order. The corresponding indices are 0, 1, 2.
   // Here we get the index of min of left and right support costs.
-  auto min_it = std::min_element(std::next(optimal_cost.begin(), 1),
-      optimal_cost.end());
-  int min_index = std::distance(optimal_cost.begin(), min_it);
+  auto min_it = std::min_element(std::next(optimal_cost->begin(), 1),
```
yangwill: This seems like overkill to get whether the left support or right support has the lower cost. An alternative could be `int min_index = optimal_cost->at(2) > optimal_cost->at(1) + 1;`. Although neither option is very readable.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -850,147 +1014,318 @@ void CassieRbtStateEstimator::contactEstimation(
...
+  double prev_t =
+      state->get_discrete_state().get_vector(time_idx_).get_value()(0);
 
   // TODO(yminchen): delete the testing code when you fix the time delay issue
   // Testing
   // current_time = cassie_out.pelvis.targetPc.taskExecutionTime;
 
   if (current_time > prev_t) {
     double dt = current_time - prev_t;
+    cout << "current_time = " << current_time << endl;
```
yangwill: this line shouldn't be here

systems/controllers/osc/operational_space_control.cc
```diff
@@ -405,20 +413,37 @@ VectorXd OperationalSpaceControl::SolveQp(
...
-        friction_constraints_.at(5 * i + 2)->UpdateLowerBound(VectorXd::Zero(1));
-        friction_constraints_.at(5 * i + 3)->UpdateLowerBound(VectorXd::Zero(1));
+        if(time_since_last_state_switch < period_of_low_friction_[i]){
+          mu_minus1 << mu_low_friction_[i], -1;
+          mu_plus1 << mu_low_friction_[i], 1;
+        } else {
+          mu_minus1 << mu_, -1;
+          mu_plus1 << mu_, 1;
+        }
+        friction_constraints_.at(5 * i)->UpdateCoefficients(
```
yangwill: Maybe add a variable with value 5 to make it clearer where the 5 comes from?

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -185,19 +185,21 @@ int DoMain(int argc, char* argv[]) {
   // important
   double w_contact_relax = 200;
   osc->SetWeightOfSoftContactConstraint(w_contact_relax);
-  // Firction coefficient
-  double mu = 0.8;
+  // Friction coefficient
+  double mu = 0.4;
+  double mu_low = 0.4;  // avoid slipping in the beginning of stance (in drake
```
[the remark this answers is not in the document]
yminchen: Removed for now. It seems to be a hack/testing code for EKF.

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -185,19 +185,21 @@ int DoMain(int argc, char* argv[]) {
...
+                        // simulation)
   osc->SetContactFriction(mu);
   Vector3d front_contact_disp(-0.0457, 0.112, 0);
   Vector3d rear_contact_disp(0.088, 0, 0);
   osc->AddStateAndContactPoint(left_stance_state,
-                               "toe_left", front_contact_disp);
+                               "toe_left", front_contact_disp, mu_low, 0.05);
```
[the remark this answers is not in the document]
yminchen: Code is removed for now.

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -263,7 +265,8 @@ int DoMain(int argc, char* argv[]) {
                                          "toe_right", "toe_rightdot");
   swing_toe_traj.AddStateAndJointToTrack(right_stance_state,
                                          "toe_left", "toe_leftdot");
-  osc->AddConstTrackingData(&swing_toe_traj, -1.5 * VectorXd::Ones(1));
+  osc->AddConstTrackingData(&swing_toe_traj, -1.5 * VectorXd::Ones(1),
```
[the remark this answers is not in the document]
yminchen: Done

systems/controllers/osc/operational_space_control.h
```diff
@@ -109,18 +109,22 @@ class OperationalSpaceControl : public drake::systems::LeafSystem<double> {
...
+                               double mu_low_friction = -1,
+                               double period_of_low_friction = 0);
 
   // Tracking data methods
   /// The third argument is used to set a period in which OSC does not track the
   /// desired traj (the period starts when the finite state machine switches to
   /// a new state)
-  void AddTrackingData(OscTrackingData* tracking_data, double duration = 0);
+  void AddTrackingData(OscTrackingData* tracking_data,
+      double t_lb = 0, double t_ub = std::numeric_limits<double>::infinity());
```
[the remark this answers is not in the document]
yminchen: OSC behaves the same in both cases. I think 0 is more informative because it tells users that the lowest lower-bound is 0.

systems/controllers/osc/operational_space_control.cc
```diff
@@ -118,22 +118,29 @@ void OperationalSpaceControl::AddAccelerationCost(std::string joint_vel_name,
 
 // Constraint methods
 void OperationalSpaceControl::AddContactPoint(std::string body_name,
-    Eigen::VectorXd pt_on_body) {
+    VectorXd pt_on_body, double mu_low_friction, double period_of_low_friction){
```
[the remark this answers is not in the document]
yminchen: Removed for now. It seems to be a hack/testing code for EKF.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -136,8 +193,8 @@ CassieRbtStateEstimator::CassieRbtStateEstimator(
         MatrixXd::Zero(6, 1), eps_cr_).
         evaluator().get();
     quadcost_eps_imu_ = quadprog_->AddQuadraticCost(MatrixXd::Zero(3, 3),
-        MatrixXd::Zero(3, 1), eps_imu_).
-        evaluator().get();
+        MatrixXd::Zero(3, 1), eps_imu_)
+        .evaluator().get();
```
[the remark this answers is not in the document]
yminchen: Done.

examples/Cassie/cassie_rbt_state_estimator.h
```diff
@@ -135,20 +170,16 @@ class CassieRbtStateEstimator : public drake::systems::LeafSystem<double> {
...
-      eps_imu_;
+  drake::solvers::VectorXDecisionVariable ddq_;
+  drake::solvers::VectorXDecisionVariable lambda_b_;
+  drake::solvers::VectorXDecisionVariable lambda_cl_;
+  drake::solvers::VectorXDecisionVariable lambda_cr_;
+  drake::solvers::VectorXDecisionVariable eps_cl_;
+  drake::solvers::VectorXDecisionVariable eps_cr_;
+  drake::solvers::VectorXDecisionVariable eps_imu_;
+  // Optimal costs
+  std::unique_ptr<std::vector<double>> optimal_cost =
```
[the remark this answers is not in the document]
yminchen: Done

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Done.

systems/controllers/osc/operational_space_control.cc
yminchen: It came from the five constraints for each contact point: 
https://github.com/DAIRLab/dairlib/blob/82df28f7f513c45ec84a41f8a0071f9b4a6a8c7c/systems/controllers/osc/operational_space_control.cc#L405-L409
I added a piece of comment in the code explaining where it comes from.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -1,5 +1,11 @@
+// TODO(yminchen): do we need to filter the gyro value before we publish?
+// We can get the bias (parameter) from EKF
+
+// TODO(yminchen): ask nanda why heel_spring_threshold_ctrl_ is 0.03?
```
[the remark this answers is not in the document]
yminchen: Done.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -816,32 +848,164 @@ void CassieRbtStateEstimator::contactEstimation(
       filtered_residual_right;
   }
 
+  // Record previous velocity (used in acceleration residual)
+  discrete_state->get_mutable_vector(
+      previous_velocity_idx_).get_mutable_value() << output.GetVelocities();
+
+  // Ground truth of contact
```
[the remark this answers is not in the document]
yminchen: Done

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Yes, moved to constructor.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Yeah I agree. I don't see the use case where we want to continue the program even though the body name doesn't exist.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Done.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Done.

examples/Cassie/cassie_rbt_state_estimator.cc
```diff
@@ -816,32 +848,164 @@ void CassieRbtStateEstimator::contactEstimation(
...
+/// EKF is updated based on the assumption of stance foot being stationary.
+/// The estimated state would get very inaccurate if the stance foot is moving.
+///
+/// Input:
+///  - OutputVector `output` containing the positions, velocities and
+///    actuator torques of the robot
+/// Ouput: left contact `left_contact` and right contact `right_contact` that
+///  indicate if the corresponding foot is in contact with the ground
+///
+/// Algorithm:
```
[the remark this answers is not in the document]
yminchen: I believe (the left spring is deflected) implies (the left-stance or double-support QP indicates contact), so I didn't add the latter into the condition. 

I will add this into comments so it's clear to the readers.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Deleted.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: Good catch. I think this might be a bug. The intension was to get the min_index among the three costs.

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: I've updated the logic so that it's using the result from double support QP

examples/Cassie/cassie_rbt_state_estimator.cc
yminchen: I didn't add optimal_cost as an input to avoid reallocating memory, but I guess it doesn't matter too much because it's only a size-3 vector.

PR #101, opened 2019-09-28
Eulerangle endeffector

makariosc:
- Cleaned up code
    - Config files now in 'config' folder, models in 'model' folder
- Added end-effector in simulation
- Added tools to zip together LCM data with ROS data

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/101)


mposa:
This PR seems really large (larger than the text implies)--is it building off of some other branch that should be committed first?

PR #104, opened 2019-10-07
Adding skeleton code for saving/loading trajectories

yangwill:
Still need to implement code marked with TODO.

Looking for feedback about overall structure before implementing the details.


yangwill:
Had some initial issues with gtest, so instead wrote a basic test file (lcm_trajectory_test). If everything looks good, I'll add a last commit to autofill metadata/write an actual unit test with gtest.

yangwill:
Forgot that I haven't implemented autofilling the git details, I'll add a commit before we merge.

lcm/BUILD.bazel
```diff
@@ -7,3 +11,26 @@ cc_binary(
         "@lcm"
     ],
 )
+
+cc_library(
+    name = "lcm_trajectory_saver",
+    srcs = ["lcm_trajectory.cc"],
+    hdrs = ["lcm_trajectory.h"],
+    deps = [
+		  "//lcmtypes:lcmt_robot",
```
mposa: Spacing looks off here

lcm/lcm_trajectory.h
```diff
@@ -0,0 +1,50 @@
...
+    std::string traj_name;
+    Eigen::VectorXd time_vector;
+    Eigen::MatrixXd datapoints;
+    std::vector<std::string> datatypes;
+  };
+
+  LcmTrajectory() {}
+  LcmTrajectory(const std::vector<Trajectory>& trajectories,
+                const std::vector<std::string>& trajectory_names,
+                const std::string& name = "DEFAULT_NAME",
```
mposa: Per f2f discussion, I think these defaults should both be empty strings.

lcm/BUILD.bazel
mposa: Spacing looks off here

lcm/lcm_trajectory.h
```diff
@@ -0,0 +1,50 @@
...
+
+  explicit LcmTrajectory(const lcmt_saved_traj& traj);
+
+  // Writes this LcmTrajectory object to a file specified by filepath
+  void writeToFile(std::string filepath);
+
+  // Loads a saved trajectory to a lcmt_saved_traj
+  static lcmt_saved_traj loadFromFile(const std::string filepath);
+
+  lcmt_metadata metadata_;
```
mposa: I doubt these properties should be public.

lcm/lcm_trajectory.h
```diff
@@ -0,0 +1,50 @@
...
+
+  LcmTrajectory() {}
+  LcmTrajectory(const std::vector<Trajectory>& trajectories,
+                const std::vector<std::string>& trajectory_names,
+                const std::string& name = "DEFAULT_NAME",
+                const std::string& description = "DEFAULT_DESCRIPTION");
+
+  explicit LcmTrajectory(const lcmt_saved_traj& traj);
+
+  // Writes this LcmTrajectory object to a file specified by filepath
```
mposa: Document what happens in the case of failure (if the file cannot be created)

lcm/lcm_trajectory.h
```diff
@@ -0,0 +1,50 @@
...
+  void writeToFile(std::string filepath);
+
+  // Loads a saved trajectory to a lcmt_saved_traj
+  static lcmt_saved_traj loadFromFile(const std::string filepath);
```
mposa: Document what happens in the case of failure (if the file cannot be opened or parsed)

lcmtypes/lcmt_saved_traj.lcm
```diff
@@ -0,0 +1,9 @@
+package dairlib;
+
+struct lcmt_saved_traj
+{
+  dairlib.lcmt_metadata metadata;
```
mposa: is `dairlib.` needed here, since you're in the same package?

lcm/lcm_trajectory.h
```diff
@@ -0,0 +1,50 @@
...
+#include <vector>
+#include <string>
+#include <unordered_map>
+#include "drake/systems/lcm/serializer.h"
+
+#include "dairlib/lcmt_saved_traj.hpp"
+
+namespace dairlib {
+
+class LcmTrajectory {
```
mposa: Please add some documentation for the class. What does it store, how should it be used, etc.

One thing that surprised me was that the row-index relates to the datapoint, rather than the trajectory index.

lcm/lcm_trajectory.cc
```diff
@@ -0,0 +1,148 @@
...
+    traj.trajectory_names.push_back(traj_el.first);
+  }
+  return traj;
+}
+
+void LcmTrajectory::writeToFile(string filepath) {
+  std::vector<uint8_t> bytes;
+  drake::systems::lcm::Serializer<lcmt_saved_traj> serializer;
+  serializer.Serialize(*AbstractValue::Make(generateLcmObject()), &bytes);
+  std::ofstream fout(filepath);
```
mposa: This might fail, setting a `failbit` (readable by the `fail` function). If it does, what should our response be? At minimum, I would think that a useful message and then throwing an error, rather than the `write` calls failing in a less interpretable manner.

lcm/lcm_trajectory.cc
```diff
@@ -0,0 +1,148 @@
...
+  fout.write(reinterpret_cast<const char*>(bytes.data()), bytes.size());
+  // fout << bytes.data();
+  fout.close();
+}
+
+lcmt_saved_traj LcmTrajectory::loadFromFile(const std::string filepath) {
+  drake::systems::lcm::Serializer<lcmt_saved_traj> serializer;
+  std::ifstream inFile(filepath, std::ios_base::binary);
```
mposa: Same as with writing. What do do here in case of failure?

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+
+TEST_F(LcmTrajectoryTest, TestConstructorFromLcmTObject) {
+  lcm_traj_.writeToFile(TEST_FILEPATH);
+
+  LcmTrajectory loaded_traj = LcmTrajectory(lcm_traj_.loadFromFile(
+                                TEST_FILEPATH));
+  EXPECT_TRUE(
+    loaded_traj.trajectories_[TEST_TRAJ_NAME_1].time_vector.isApprox(
+      lcm_traj_.trajectories_[TEST_TRAJ_NAME_1].time_vector));
+  EXPECT_TRUE(
```
mposa: We should probably also check that the other elements also check out after the save-load process.

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+    loaded_traj.trajectories_[TEST_TRAJ_NAME_1].time_vector.isApprox(
```
mposa: Looks like a duplicate check on `TEST_TRAJ_NAME_1`. Should this be `TEST_TRAJ_NAME_2`?

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+      lcm_traj_.trajectories_[TEST_TRAJ_NAME_1].time_vector));
+}
```
mposa: Do you have any idea if bazel will automatically clean up the file you're creating here? If it doesn't, should we delete it?

lcm/lcm_trajectory.h
yangwill: I'm open to using getters/setters, but I don't think they add much utility in this case. Unless you are referring to something else.

mposa: Would setters even be needed? One reason for making private is that `trajectories_` and `trajectory_names_` cannot be changed independently.

lcm/lcm_trajectory.h
mshalm: I would say in most software packages I've used, vector-valued time series data uses this format (rows ~ time indexing), but given that we usually think of state as a column vector I would say documenting is definitely in order.

mshalm: As for performance, it seems that the _rows_ are getting compact memory allocation, and memory operations are being done on a per-row basis. It would likely be more efficient to flip the rows/columns, but I don't know that we'll be handling data in quantities for which that will make much of a difference.

mshalm:
Looks great to me modulo minor comments!

lcmtypes/lcmt_metadata.lcm
```diff
@@ -0,0 +1,10 @@
+package dairlib;
+
+struct lcmt_metadata 
+{
+  boolean git_dirty_flag;
```
yminchen: Adding a comment on what `git_dirty_flag` is should be helpful

lcmtypes/lcmt_trajectory_block.lcm
```diff
@@ -0,0 +1,12 @@
+package dairlib;
+
+struct lcmt_trajectory_block
+{
+  string trajectory_name;
+  
+  int32_t num_points;
+  int32_t num_datatypes;
+  double time_vec[num_points];
+  double datapoints[num_points][num_datatypes];
```
yminchen: It will be helpful to add comments on what `dataypes` is here, since I'm not very sure what the second dimension of datapoints is here. Maybe the name `datatypes` confuses me. 
I guess `datapoints` stores a `num_datatypes`-size vector v(t)? For example, an input u(t) with n_u number of actuation.

lcm/lcm_trajectory.cc
```diff
@@ -0,0 +1,148 @@
...
+                         static_cast<int>(bytes.size()),
+                         traj_value.get());
+  return traj_value->get_value<lcmt_saved_traj>();
+}
+
+lcmt_metadata LcmTrajectory::constructMetadataObject(string name,
+    string description) const {
+  lcmt_metadata metadata;
+
+  metadata.datetime = "";
```
yminchen: `datetime` is not set?

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+static const char DATATYPE_2[] = "DATATYPE_2";
+static const int NUM_DATAPOINTS = 5;
+static const int NUM_DATATYPES = 2;
+static const int NUM_TRAJECTORIES = 2;
+
+
+class LcmTrajectoryTest : public ::testing::Test {
+ protected:
+  void SetUp() override {
+    datatypes_.resize(NUM_DATAPOINTS);
```
yminchen: should this be `datatypes_.resize(NUM_DATATYPES);`?

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+    datatypes_[0] = DATATYPE_1;
+    datatypes_[1] = DATATYPE_2;
+
+    traj_1_.traj_name = TEST_TRAJ_NAME_1;
+    traj_1_.time_vector = VectorXd::Ones(NUM_DATAPOINTS);
+    traj_1_.datapoints = MatrixXd::Identity(NUM_DATAPOINTS, NUM_DATAPOINTS);
```
yminchen: Should this be `MatrixXd::Identity(NUM_DATATYPES, NUM_DATAPOINTS);` or `MatrixXd::Identity(NUM_DATAPOINTS, NUM_DATATYPES);` ?

lcm/test/lcm_trajectory_test.cc
yminchen: @yangwill FYI, if you need to delete the file manually, you can refer to my code here: https://github.com/DAIRLab/dairlib/blob/goldilocks-model-dev/examples/goldilocks_models/find_goldilocks_models.cc#L452

lcm/test/lcm_trajectory_test.cc
```diff
@@ -0,0 +1,122 @@
...
+
+    lcm_traj_ = LcmTrajectory(trajectories_,
+                              trajectory_names_,
+                              TEST_NAME,
+                              TEST_DESCRIPTION);
+  }
+
+  lcmt_trajectory_block traj_block_1_;
+  lcmt_trajectory_block traj_block_2_;
+  lcmt_saved_traj lcmt_traj_;
```
yminchen: Looks like `traj_block_1_`, `traj_block_2_` and `lcmt_traj_` are not used here, so we can delete them.

lcm/lcm_trajectory.h
yminchen: Just adding my opinion here. I'm using to columns as the time index, and all of my code follows this rule. It will be great if we use columns for time.

Drake's cubic PiecewisePolynomial also follows the same rule 
https://drake.mit.edu/doxygen_cxx/classdrake_1_1trajectories_1_1_piecewise_polynomial.html#a81996df67cccc58f69bcff1f3e6b9671

lcm/test/lcm_trajectory_test.cc
yangwill: I wasn't able to find whether or not bazel test would automatically delete the files. So instead, I searched for the files create from the test using find . "TEST_FILEPATH" and couldn't find anything. I'm not sure if that's a satisfying answer.

lcmtypes/lcmt_trajectory_block.lcm
yangwill: You shouldn't have to interface with this lcm object, I'm using it as a convenience to store/load using the LCM serializer. Although that's not the case for lcmt_metadata object.

lcm/lcm_trajectory.h
yangwill: Changed datapoints so that the datatype is indexed by row, and times are indexed by column. Because I don't anticipate calling a save/load often per program or during a time-critical operation, I don't think it matters how we store the memory.

lcm/lcm_trajectory.cc
```diff
@@ -68,9 +79,9 @@ lcmt_saved_traj LcmTrajectory::generateLcmObject() const {
   traj.num_trajectories = trajectories_.size();
 
   // For each trajectory
-  for (auto & traj_el : trajectories_) {
+  for (auto& traj_el : trajectories_) {
     lcmt_trajectory_block traj_block;
-    const Trajectory* cpp_traj = &traj_el.second;
+    const Trajectory *cpp_traj = &traj_el.second;
```
mposa: Per GSG, this should be `Trajectory* cpp_traj` not `Trajectory *cpp_traj` (the way your earlier commit had it)

lcm/lcm_trajectory.h
mposa: The header for these functions should indicate that an exception will be thrown in case of an error.

lcmtypes/lcmt_saved_traj.lcm
mposa: Bumping this older comment.

PR #107, opened 2019-10-21
Add benchmark example for dircon

yminchen:
Fix bugs in dircon and add a benchmark example for dircon's speed. 

**Detail changes**:
Dircon:
- fix a bug. Use the jacobian that doesn't skip any constraints in the equations of motion
- fix bugs in the index of decision variable when adding constraint 
- pass in surface normal into constructor of dircon_position_data
- add surface (e.g. ground) incline feature to dircon_position_data
- use linear constraint instead of LorentzConeConstraint for friction cone (to increase speed)
- add a utility function to detect if a MBP is quaternion floating based
- add unit norm quaternion constraint and also add a slack variable in dynamic constraint 

Example:
- Trajectory optimization of Cassie squating

### DIFF
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -22,17 +22,12 @@
 #include "drake/solvers/constraint.h"
 #include "drake/solvers/choose_best_solver.h"
 
-#include "drake/multibody/inverse_kinematics/inverse_kinematics.h"
-
 #include "drake/multibody/rigid_body_tree_construction.h"
 #include "drake/multibody/rigid_body_tree.h"
 #include "drake/multibody/rigid_body_plant/drake_visualizer.h"
-#include "drake/systems/primitives/trajectory_source.h"
 
 #include "common/find_resource.h"
-#include "systems/primitives/subvector_pass_through.h"
 
-#include "solvers/optimization_utils.h"
 #include "systems/trajectory_optimization/dircon_position_data.h"
 #include "systems/trajectory_optimization/dircon_distance_data.h"
 #include "systems/trajectory_optimization/dircon_kinematic_data_set.h"
@@ -40,9 +35,7 @@
 #include "systems/trajectory_optimization/hybrid_dircon.h"
 
 #include "multibody/multibody_utils.h"
-#include "multibody/visualization_utils.h"
 
-#include "systems/goldilocks_models/symbolic_manifold.h"
 #include "systems/goldilocks_models/file_utils.h"
 
 #include "examples/Cassie/cassie_utils.h"
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -122,13 +122,13 @@ DEFINE_double(tol, 1e-4,
 
 namespace dairlib {
 
-/// Trajectory optimization of fixed-spring cassie squating
+/// Trajectory optimization of fixed-spring cassie squatting
 /// With the default initial guess, the solving time is about 100 seconds.
 
 // Constraint to fix the position of a point on a body (for initial guess)
-class BodyPonitPositionConstraint : public DirconAbstractConstraint<double> {
+class BodyPointPositionConstraint : public DirconAbstractConstraint<double> {
  public:
-  BodyPonitPositionConstraint(const RigidBodyTree<double>& tree,
+  BodyPointPositionConstraint(const RigidBodyTree<double>& tree,
                               string body_name,
                               Vector3d translation,
                               Vector3d desired_pos) :
@@ -142,7 +142,7 @@ class BodyPonitPositionConstraint : public DirconAbstractConstraint<double> {
     translation_(translation),
     desired_pos_(desired_pos) {
   }
-  ~BodyPonitPositionConstraint() override = default;
+  ~BodyPointPositionConstraint() override = default;
 
   void EvaluateConstraint(const Eigen::Ref<const drake::VectorX<double>>& x,
                           drake::VectorX<double>* y) const override {
@@ -159,7 +159,7 @@ class BodyPonitPositionConstraint : public DirconAbstractConstraint<double> {
   const Vector3d desired_pos_;
 };
 
-// Use fixed-point solver to get configuration guess (for standing in place)
+// Use fixed-point solver to get initial guess
 void GetInitFixedPointGuess(const Vector3d& pelvis_position,
                             const RigidBodyTree<double>& tree,
                             VectorXd* q_init,
@@ -186,35 +186,39 @@ void GetInitFixedPointGuess(const Vector3d& pelvis_position,
   ContactInfo contact_info(xa, idxa);
 
   VectorXd q_desired = VectorXd::Zero(n_q);
-  q_desired << -0.0872062,
-            1.56883E-13,
-            1,
-            1,
-            1.9884E-13,
-            1.30167E-14,
-            -2.10728E-14,
-            -0.0104994,
-            0.0104994,
-            -3.42713E-07,
-            -3.30511E-07,
-            0.509601,
-            0.509602,
-            -1.2219,
-            -1.22191,
-            1.44602,
-            1.44602,
-            -1.6072,
-            -1.6072;
+  q_desired <<         0,
+                       0,
+                   1.057,
+                       1,
+                       0,
+                       0,
+                       0,
+                  0.0185,
+                 -0.0185,
+                       0,
+                       0,
+                   0.383,
+                   0.383,
+                   -1.02,
+                   -1.02,
+                    1.24,
+                    1.24,
+                   -1.48,
+                   -1.48;
+  // The above numbers comes from one (FixedPointSolver) solution of cassie
+  // standing
 
   std::map<int, double> fixed_joints;
+  // floating base x, y z position
   for (int i = 0; i < 3; i++) {
     fixed_joints[i] = pelvis_position[i];
   }
+  // floating base quaternion
   fixed_joints[3] = 1;
   fixed_joints[4] = 0;
   fixed_joints[5] = 0;
   fixed_joints[6] = 0;
-
+  // hip yaw position
   fixed_joints[9] = 0;
   fixed_joints[10] = 0;
 
@@ -231,12 +235,12 @@ void GetInitFixedPointGuess(const Vector3d& pelvis_position,
   shared_ptr<MathematicalProgram> mp = fp_solver.get_program();
   auto& q_var = mp->decision_variables().head(
                   n_q);  // Assume q is located at the start
-  Vector3d desired_left_toe_pos(0.05, 0.15, 0);
-  Vector3d desired_right_toe_pos(0.05, -0.15, 0);
-  auto left_foot_constraint = std::make_shared<BodyPonitPositionConstraint>(
+  Vector3d desired_left_toe_pos(0.06, 0.15, 0);
+  Vector3d desired_right_toe_pos(0.06, -0.15, 0);
+  auto left_foot_constraint = std::make_shared<BodyPointPositionConstraint>(
                                 tree, "toe_left", pt_front_contact,
                                 desired_left_toe_pos);
-  auto right_foot_constraint = std::make_shared<BodyPonitPositionConstraint>(
+  auto right_foot_constraint = std::make_shared<BodyPointPositionConstraint>(
                                  tree, "toe_right", pt_front_contact,
                                  desired_right_toe_pos);
   mp->AddConstraint(left_foot_constraint, q_var);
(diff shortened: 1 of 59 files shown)


examples/Cassie/run_dircon_squatting.cc
```diff
@@ -0,0 +1,817 @@
...
+  xa.col(3) = pt_rear_contact;
+  std::vector<int> idxa;
+  idxa.push_back(toe_left_idx);
+  idxa.push_back(toe_left_idx);
+  idxa.push_back(toe_right_idx);
+  idxa.push_back(toe_right_idx);
+  ContactInfo contact_info(xa, idxa);
+
+  VectorXd q_desired = VectorXd::Zero(n_q);
+  q_desired << -0.0872062,
```
mposa: Where did these magic numbers come from (comment would be useful). Can these be rounded?

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -0,0 +1,817 @@
...
+            1.44602,
+            1.44602,
+            -1.6072,
+            -1.6072;
+
+  std::map<int, double> fixed_joints;
+  for (int i = 0; i < 3; i++) {
+    fixed_joints[i] = pelvis_position[i];
+  }
+  fixed_joints[3] = 1;
```
mposa: Which joints are these?

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -70,39 +73,44 @@ HybridDircon<T>::HybridDircon(
...
-    DRAKE_ASSERT(static_cast<int>(constraint->num_constraints()) == num_states());
+    // slack variable for unit norm constraint of quaternion floating-base
+    int num_quat_slack = (multibody::isQuaternion(plant))? 1 : 0;
 
     // For N-1 timesteps, add a constraint which depends on the knot
     // value along with the state and input vectors at that knot and the
     // next.
 
     //Adding dynamic constraints
+    auto constraint = std::make_shared<DirconDynamicConstraint<T>>(plant_, *constraints_[i], num_quat_slack);
```
mposa: Shouldn't the quaternion constraint be created and added here, rather than in the main file?

systems/trajectory_optimization/dircon_kinematic_data_set.cc
```diff
@@ -109,7 +110,7 @@ void DirconKinematicDataSet<T>::updateData(const Context<T>& context,
     right_hand_side_ = -right_hand_side_ +
         plant_.MakeActuationMatrix() * input +
         plant_.CalcGravityGeneralizedForces(context) +
-        getJ().transpose() * forces;
+        J_.transpose() * forces;
```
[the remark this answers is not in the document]
yminchen: Yes, `getJWithoutSkipping()` is probably better here

systems/trajectory_optimization/dircon_opt_constraints.h
```diff
@@ -69,15 +70,17 @@ class DirconDynamicConstraint : public DirconAbstractConstraint<T> {
   DirconDynamicConstraint(const drake::multibody::MultibodyPlant<T>& plant,
                           DirconKinematicDataSet<T>& constraints,
                           int num_positions, int num_velocities, int num_inputs,
-                          int num_kinematic_constraints);
+                          int num_kinematic_constraints_wo_skipping,
+                          int num_quat_slack);
```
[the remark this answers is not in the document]
yminchen: `num_quat_slack` is the dimension of the slack variable for quaternion. It's always 1 if the MBP is quaternion floating based. Otherwise, it's 0. I'll add documentation about this.

systems/trajectory_optimization/dircon_opt_constraints.cc
```diff
@@ -176,8 +183,15 @@ void DirconDynamicConstraint<T>::EvaluateConstraint(
   auto g = constraints_->getXDot();
   VectorX<T> vc_in_qdot_space(num_positions_);
   plant_.MapVelocityToQDot(*contextcol,
-      constraints_->getJ().transpose()*vc, &vc_in_qdot_space);
+      constraints_->getJWithoutSkipping().transpose()*vc, &vc_in_qdot_space);
   g.head(num_positions_) += vc_in_qdot_space;
+
+  // The slack variable allows the quaternion to stay on a unit sphere.
+  if (num_quat_slack_ > 0) {
```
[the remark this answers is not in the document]
yminchen: I added documentation for `num_quat_slack` in the class `DirconDynamicConstraint`.

Assuming the floating base is in the first four element speeds up the computation. I just added a check in the constructor to ensure the first four elements correspond to quaternion.

examples/Cassie/run_dircon_squatting.cc
yminchen: It came from inverse kinematics where cassie was standing on the ground. 
I also added documentation and rounded some of the numbers.

examples/Cassie/run_dircon_squatting.cc
yminchen: 0 to 2 is base x y z, 3 to 6 is quaternion, and 9 and 10 is hip yaw. (documented in new commits)

yminchen:
@mposa  I updated the code according to the review

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: Yes. I made this change in the new commits

systems/trajectory_optimization/dircon_position_data.h
```diff
@@ -13,7 +13,8 @@ class DirconPositionData : public DirconKinematicData<T> {
  public:
   DirconPositionData(const drake::multibody::MultibodyPlant<T>& plant,
     const drake::multibody::Body<T>& body,
-    Eigen::Vector3d pt, bool isXZ = false);
+    Eigen::Vector3d pt, bool isXZ = false,
+    Eigen::Vector2d surface_slope_roll_pitch = Eigen::Vector2d::Zero());
```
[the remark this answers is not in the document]
yminchen: I agree. I replaced roll and pitch with normal.

systems/trajectory_optimization/dircon_position_data.cc
```diff
@@ -18,13 +20,21 @@ using drake::MatrixX;
...
+    Vector3d surface_normal) :
+        DirconKinematicData<T>(plant, isXZ ? 2 : 3),
+        body_(body),
+        pt_(pt),
+        isXZ_(isXZ) {
   TXZ_ << 1, 0, 0,
           0, 0, 1;
+
+  Vector3d z_hat(0,0,1);
+  surface_normal.normalize();
```
mposa: Does it make sense to supply a `surface_normal` value if `isXZ = true`? If these two arguments cannot be supplied together, then this should either be prevented via the API or, at minimum, checked at runtime.

systems/trajectory_optimization/dircon_position_data.cc
```diff
@@ -56,45 +66,56 @@ void DirconPositionData<T>::updateConstraint(const Context<T>& context) {
...
-                                                              double mu) {
+void DirconPositionData<T>::addFixedNormalFrictionConstraints(double mu) {
   if (isXZ_) {
-    // specifically builds the basis for the x-axis
-    Vector2d normal_xz, d_xz;
-    double L = sqrt(normal(0)*normal(0) + normal(2)*normal(2));
-    normal_xz << normal(0)/L, normal(2)/L;
-    d_xz << -normal_xz(1), normal_xz(0);
+    // builds the basis
+    Vector2d normal_xz(0,1);
```
mposa: Related to earlier comment re: normal and isXZ. This looks like a change where, previously, we supported non-flat normal vectors for planar contact. Is this no longer the case? If not, why are we removing this functionality?

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -24,34 +22,39 @@ using drake::systems::trajectory_optimization::MultipleShooting;
...
-    mode_lengths_(num_time_samples),
-    v_post_impact_vars_(NewContinuousVariables(plant.num_velocities() *
-                        (num_time_samples.size() - 1), "v_p")) {
+HybridDircon<T>::HybridDircon(const MultibodyPlant<T>& plant,
+                              vector<int> num_time_samples,
+                              vector<double> minimum_timestep,
+                              vector<double> maximum_timestep,
+                              vector<DirconKinematicDataSet<T>*> constraints,
+                              vector<DirconOptions> options)
+    : MultipleShooting(plant.num_actuators(),
```
mposa: Spacing looks off here.

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -24,34 +22,39 @@ using drake::systems::trajectory_optimization::MultipleShooting;
...
+                                       num_time_samples.end(),
+                                       0) - num_time_samples.size() + 1,
+                       1e-8,
+                       1e8),
+      plant_(plant),
+      constraints_(constraints),
+      num_modes_(num_time_samples.size()),
+      mode_lengths_(num_time_samples),
+      v_post_impact_vars_(NewContinuousVariables(
+          plant.num_velocities() *
```
mposa: Also here

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -66,118 +69,187 @@ HybridDircon<T>::HybridDircon(
...
+            * num_time_samples[i], "lambda[" + std::to_string(i) + "]"));
+    collocation_force_vars_.push_back(NewContinuousVariables(
+        constraints_[i]->countConstraintsWithoutSkipping()
+            * (num_time_samples[i] - 1),
+        "lambda_c[" + std::to_string(i) + "]"));
+    collocation_slack_vars_.push_back(NewContinuousVariables(
+        constraints_[i]->countConstraintsWithoutSkipping()
+            * (num_time_samples[i] - 1), "v_c[" + std::to_string(i) + "]"));
+    // slack variables used to scale quaternion norm to 1 in the dynamic constraints.
+    quaternion_slack_vars_.push_back(NewContinuousVariables(
```
mposa: This isn't very readable. Why not
```
if (is_quaternion) {
// Add variables

}
```

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -66,118 +69,187 @@ HybridDircon<T>::HybridDircon(
...
-
-    DRAKE_ASSERT(static_cast<int>(constraint->num_constraints()) == num_states());
 
     // For N-1 timesteps, add a constraint which depends on the knot
     // value along with the state and input vectors at that knot and the
     // next.
 
+
+    //Adding quaternion norm constraint
+    if (multibody::isQuaternion(plant)) {
```
mposa: Should use the boolean you've created, `is_quaternion` here.

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -66,118 +69,187 @@ HybridDircon<T>::HybridDircon(
...
+      auto quat_norm_constraint =
+          std::make_shared<QuaternionNormConstraint<T>>();
+      for (int j = (i == 0) ? 0 : 1; j < mode_lengths_[i]; j++) {
```
mposa: Add a comment here to explain the use of `x ? y : z`.

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: I couldn't tell how it is off. I used a code style formatter to format `hybrid_dircon.cc`.
The style should be the same (if I didn't miss any thing) as Drake's. https://github.com/RobotLocomotion/drake/blob/master/systems/trajectory_optimization/multiple_shooting.cc#L63-L92

systems/trajectory_optimization/dircon_opt_constraints.cc
```diff
@@ -107,29 +110,57 @@ void DirconAbstractConstraint<double>::DoEval(
 }
 
 
+template <typename T>
+QuaternionNormConstraint<T>::QuaternionNormConstraint() :
+    DirconAbstractConstraint<T>(1, 4,
+                                       VectorXd::Zero(1),
```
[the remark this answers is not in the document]
yminchen: Yes, I forgot to update the code spacing of this file

systems/trajectory_optimization/dircon_position_data.cc
```diff
@@ -18,13 +20,21 @@ using drake::MatrixX;
...
 template <typename T>
 DirconPositionData<T>::DirconPositionData(const MultibodyPlant<T>& plant,
-    const Body<T>& body, Vector3d pt, bool isXZ) :
-    DirconKinematicData<T>(plant, isXZ ? 2 : 3),
-    body_(body),
-    pt_(pt),
-    isXZ_(isXZ) {
+    const Body<T>& body, Vector3d pt, bool isXZ,
+    Vector3d surface_normal) :
+        DirconKinematicData<T>(plant, isXZ ? 2 : 3),
```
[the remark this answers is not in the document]
yminchen: Yes, I'll update the code style of this file

systems/trajectory_optimization/dircon_position_data.cc
yminchen: Yes, I use both of them in the example of planar five-link robot example.

systems/trajectory_optimization/dircon_position_data.cc
yminchen: Currently there are c, cdot and cddot in `dircon_kinemaitcs_data`. We want c=0 in the normal direction of the surface, so we rotate the contact position into the surface frame. (I guess we can generalize this more by adding a translation, if there is a specific position where we want the robot to touch the surface? Currently the surface has to contain the global origin.)

Since we rotate c, cdot and cddot because of the above reason, we don't need to rotate it again in the `addFixedNormalFrictionConstraints()`.

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: Same comment. I couldn't tell how it is off. I think this is the same as drake's style?

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: I feel they are equally readable, but I will change to 
```
if (is_quaternion) {
// Add variables

}
```

yminchen:
@mposa I've added comments and updated code spacing, but I'm not sure if this is the code style you want.

mposa:
Per face to face with Yu-Ming, add an option to fix the seed for the RNG so runtime comparisons are consistent.

PR #108, opened 2019-10-29
Added Cassie joint velocity shutdown

mposa:
Added a new input supervisor. Right now, this serves to shut down our controller whenever the velocity of the robot exceeds a threshold. Copied from the class  header below. I've tested this briefly against the Gazebo simulation.

/// The InputSupervisor is a simple Drake System that acts as an intermediary
/// between commands from controllers and the actual robot. It's envisioned role
/// is to (1) sanity check any inputs, (2) shut down in case of errors or
/// instability, and (3) to mediate between multiple controllers. Of these three
/// potential purposes, only (2) is currently implemented as a check on the
/// current velocity of the robot. If the velocity passes a specified threshold,
/// in absolute value, then an error flag (discrete state) is set and the output
/// commands are set to zero. Note, in the presence of noise, this __might__ be
/// a brittle check. If this is the case, we will have to low-pass filter this
/// check. For instance, we could only set the flag if the threshold is reached
/// N times in a row.

### DIFF
--- a/examples/Cassie/dispatcher_robot_in.cc
+++ b/examples/Cassie/dispatcher_robot_in.cc
@@ -1,4 +1,5 @@
 #include <memory>
+#include <limits>
 
 #include <gflags/gflags.h>
 #include "drake/lcm/drake_lcm.h"
@@ -31,7 +32,11 @@ using drake::systems::TriggerType;
 DEFINE_string(address, "127.0.0.1", "IPv4 address to publish to (UDP).");
 DEFINE_int64(port, 25000, "Port to publish to (UDP).");
 DEFINE_double(pub_rate, .02, "Network LCM pubishing period (s).");
-DEFINE_double(max_joint_velocity, 10, "Maximum joint velocity before error is triggered");
+DEFINE_double(max_joint_velocity, 10,
+    "Maximum joint velocity before error is triggered");
+DEFINE_double(input_limit, -1, "Maximum torque limit. Negative values are inf.");
+DEFINE_int64(supervisor_N, 10,
+    "Maximum allowed consecutive failures of velocity limit.");
 DEFINE_string(state_channel_name, "CASSIE_STATE",
     "The name of the lcm channel that sends Cassie's state");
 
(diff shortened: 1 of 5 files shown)


yangwill:
I with Yuming that we might want to add additional checks such as joint/torque limits in future commits. This looks good for now. We can add a check for N consecutive threshold violations if the velocity shutdown is too sensitive during testing.

mposa:
It's easy enough to add code for `N` consecutive failures, and just set `N=1` for now. I'll make this change and then merge if it looks good to you.

examples/Cassie/dispatcher_robot_in.cc
```diff
@@ -56,15 +60,25 @@ int do_main(int argc, char* argv[]) {
...
+  builder.Connect(state_receiver->get_output_port(0),
+      input_supervisor->get_input_port_state());
+  builder.Connect(command_receiver->get_output_port(0),
+      input_supervisor->get_input_port_command());
 
   // Create and connect translator
   auto input_translator =
       builder.AddSystem<systems::CassieInputTranslator>(*tree);
-  builder.Connect(*command_receiver, *input_translator);
+  builder.Connect(*input_supervisor, *input_translator);
```
yminchen: In line 96, should it be 
`builder.Connect(*input_supervisor, *net_command_sender);`
instead of 
`builder.Connect(*command_receiver, *net_command_sender);`?

The command after the guard is probably what we want to see on the Sherriff end.

mposa:
> Another idea--I think we can also have a flag to set an input bound manually. There are some cases where we know only little torque is needed (for example when we test the standing controller) We can manually cap the input to be less than 100 to protect the robot. Not sure how valuable this is.

Good idea. Will add.

mposa:
@yminchen @yangwill I added a input saturation mechanism and the N-consecutive failures. Please re-review.

yminchen:
:lgtm:

yangwill:
Looks good to me. There might be a point in the future where we want to output if the input supervisor  is in shutdown mode, but we can add that later.

PR #109, opened 2019-11-03
Osc update

yminchen:
There is a critical bug in OSC that was fixed in `invariant-ekf-yuming` which is not merged to master yet. This PR fixes the bug. 

I also updated gitignore to ignore the folder that is generated by CLion


yminchen:
There are other changes that's not osc related. Will re-create PR.

PR #112, opened 2019-11-08
Undelete LcmMultiplexer

yangwill:
LcmMultiplexer was accidentally deleted in an old commit. This commit re-adds the LcmMultiplexer and updates the test to work with the new drake deprecations.


mposa:
Looks like it's failing tests because `std::optional` was only introduced in C++17 (which we haven't upgraded to yet)

yangwill:
> Looks like it's failing tests because `std::optional` was only introduced in C++17 (which we haven't upgraded to yet). `drake::optional` replicates the functionality (I think). Not totally sure why this fails now, and didn't before.

Drake is deprecating drake::optional. The reason it fails now is because we had to replace the previous mocked function with the more recent Publish() function that requested an optional. I'll change it to drake::optional.

mposa:
Is this superseded by #114 ?

yangwill:
> Is this superseded by #114 ?

Yeah. The LCM mux will implemented in #114 instead.

PR #114, opened 2019-11-18
Add driven loop mux

yminchen:
- Add `LcmDrivenLoop` class which has MUX feature
- Update the scripts of cassie osc controllers and dispatchers to use `LcmDrivenLoop`
- Add `ControllerChannelSender` to publish lcm message that contains channel name 

The classes were tested by running the scripts `run_osc_standing_controller`, `run_controller_switch`, `dispatch_robot_in`, `run_simple_sim`, `visualizer` and also lcm-spy.


yminchen:
@mposa The PR is ready for review

MUJOCO_LOG.TXT
```diff
@@ -0,0 +1,3 @@
+Fri Nov 22 14:32:28 2019
```
mposa: Don't think this file should be added.

.clang-format
```diff
@@ -0,0 +1,45 @@
+# -*- yaml -*-
```
mposa: Can you merge with master? I think this is a duplicate addition.

systems/framework/BUILD.bazel
```diff
@@ -48,7 +48,6 @@ cc_test(
 cc_library(
```
mposa: Entire binary should be deleted.

systems/framework/lcm_driven_loop.h
```diff
@@ -4,90 +4,207 @@
...
+  ///     @param lcm_parser The LeafSystem of the diagram that parses the
+  ///     incoming lcm message
+  ///     @param input_channel The name of the input channel
   LcmDrivenLoop(drake::lcm::DrakeLcm* drake_lcm,
                 std::unique_ptr<drake::systems::Diagram<double>> diagram,
-                const drake::systems::LeafSystem<double>* first_leafsystem,
-                const std::string& input_channel);
-
-  // Lcm driven loop for multiple inputs
+                const drake::systems::LeafSystem<double>* lcm_parser,
```
mposa: Can/should this just call the multi-channel constructor (with just a length 1 list)? If so, I don't see a need for the private constructor.

systems/framework/lcm_driven_loop.h
```diff
@@ -4,90 +4,207 @@
...
+            std::cout << switch_sub_->message().channel << " doesn't exist\n";
+          }
+          switch_sub_->clear();
+        }
+      }
+
+      // Wait for an InputMessageType message.
+      name_to_input_sub_map_.at(active_channel_).clear();
+      LcmHandleSubscriptionsUntil(drake_lcm_, [&]() {
+        return (name_to_input_sub_map_.at(active_channel_).count() > 0);
```
mposa: Shouldn't this exit on either the `active_channel` subscriber or the switching subscriber, with a different result depending on which is triggered?

MUJOCO_LOG.TXT
yminchen: I added accidentally. Just deleted the file.

.clang-format
yminchen: Done

systems/framework/BUILD.bazel
yminchen: What do you mean by deleting the binary?

systems/framework/lcm_driven_loop.h
yminchen: Yes, we can have one multi-channel constructor only, but I think it is unnatural to create a length 1 list for the single input channel

yminchen: I will change to single constructor for shorter code anyways.

systems/framework/lcm_driven_loop.h
yminchen: I thought about that, but it makes the logic more complicated (e.g. we need a variable to remember the previous message), and do you think it's necessary?

systems/framework/BUILD.bazel
mposa: My mistake--I misinterpreted something.

systems/framework/lcm_driven_loop.h
mposa: I think having two constructors is good, but I was questioning the need for the first. I think something like:

```
  LcmDrivenLoop(drake::lcm::DrakeLcm* drake_lcm,
                std::unique_ptr<drake::systems::Diagram<double>> diagram,
                const drake::systems::LeafSystem<double>* lcm_parser,
                const std::string& input_channel)
      : LcmDrivenLoop(drake_lcm, std::move(diagram), input_channel, "", std::vector<std::string>{input_channel}, lcm_parser) {}
```

where the single-channel constructor just calls the multi-channel constructor (with a blank switching channel argument and length 1 vector)

systems/framework/lcm_driven_loop.h
mposa: I think it's a big advantage, as it will allow us to potentially recover from controllers that crash.

Why does it need to remember the previous message?

systems/framework/lcm_driven_loop.h
```diff
@@ -0,0 +1,217 @@
...
+
+    // "Simulator" time
+    double time = 0;  // initialize the current time with 0
+    double message_time;
+
+    drake::log()->info(diagram_name_ + " started");
+    // Run the simulation until end_time
+    while (time < end_time) {
+      // Update the name of the active channel if there are multiple inputs
+      if (is_multiple_inputs_) {
```
mposa: Can this just check `switch_sub_ == nullptr`?

yminchen: Yes, it's done

systems/framework/lcm_driven_loop.h
yminchen: Ok, the feature is added. I was wrong that we need to remember the previous message. In the implementation, I just don't update the message value when there is no new input message.

yminchen:
@mposa The PR is ready for review

systems/framework/lcm_driven_loop.h
```diff
@@ -148,14 +159,26 @@ class LcmDrivenLoop {
...
+        return is_new_input_message || is_new_switch_message;
       });
 
       // Write the InputMessageType message into the context.
-      input_value.GetMutableData()->set_value(
-          name_to_input_sub_map_.at(active_channel_).message());
+      if (is_new_input_message) {
+        input_value.GetMutableData()->set_value(
+            name_to_input_sub_map_.at(active_channel_).message());
+      }
```
mposa: Should the entire block of code below (where the simulator is advanced) also be inside this `if` statement?

yminchen: Yes, it's done!

systems/framework/lcm_driven_loop.h
```diff
@@ -0,0 +1,227 @@
...
+
+    // "Simulator" time
+    double time = 0;  // initialize the current time with 0
+    double message_time;
+
+    drake::log()->info(diagram_name_ + " started");
+    // Run the simulation until end_time
+    while (time < end_time) {
+      // Update the name of the active channel if there are multiple inputs
+      if (switch_sub_ != nullptr) {
```
mposa: One (hopefully last) comment. For better readability, with the current structure, it feels like this code (executing the channel switch) should come after the message is received (rather than before).

I know it doesn't matter much, but I think it's easier to understand if the code looks like:
```
while true
   wait for message A or B
   if A, change channel
   if B, step simulation
```
rather than
```
while true
   if A, change channel
   wait for message A or B
   if B, step simulation
```

yminchen: Done. I also updated the logic of clearing messages, so now we clear the new channel messages during the loop when we just switch to a new channel, in case we read the old messages.

systems/framework/lcm_driven_loop.h
```diff
@@ -198,14 +196,41 @@ class LcmDrivenLoop {
...
+        }
+
+        // Clear messages in the switch channel
+        name_to_input_sub_map_.at(active_channel_).clear();
+        switch_sub_->clear();
+
+        // Clear messages in the new input channel if we just switched input
+        // channel in the current loop
+        if (previous_active_channel_name.compare(active_channel_) != 0) {
+          name_to_input_sub_map_.at(active_channel_).clear();
```
mposa: Doesn't seem like there's a need for this second `.clear()` call here. I do think it's always OK to clear the new active channel (line 224), so probably just need to remove 229-231.

yminchen: Line 224 was unintentional, and I've removed it. So the idea is that we only clear the input channel after 
1. we read the input messgae
2. we just switch to a new input channel

PR #117, opened 2019-11-25
Adding class that allows user to simulate sensor drift

yangwill:
Adding leaf system that is intended to be between the controller's state_receiver and controller input port. It allows the user to specify the drift as a Gaussian distribution.

### DIFF
--- a/examples/Cassie/simulator_drift.cc
+++ b/examples/Cassie/simulator_drift.cc
@@ -1,7 +1,3 @@
-//
-// Created by yangwill on 11/15/19.
-//
-
 #include "simulator_drift.h"
 
 using dairlib::systems::OutputVector;
--- a/examples/Cassie/simulator_drift.cc
+++ b/examples/Cassie/simulator_drift.cc
@@ -1,0 +1,68 @@
+//
+// Created by yangwill on 11/15/19.
+//
+
+#include "simulator_drift.h"
+
+using dairlib::systems::OutputVector;
+using Eigen::VectorXd;
+
+SimulatorDrift::SimulatorDrift(const RigidBodyTree<double>& tree,
+                               const Eigen::VectorXd& drift_mean,
+                               const Eigen::MatrixXd& drift_cov)
+    : tree_(tree), drift_mean_(drift_mean), drift_cov_(drift_cov) {
+  DRAKE_ASSERT(drift_mean_.size() == drift_cov_.cols());
+  state_port_ = this
+                    ->DeclareVectorInputPort(OutputVector<double>(
+                        tree_.get_num_positions(), tree_.get_num_velocities(),
+                        tree_.get_num_actuators()))
+                    .get_index();
+
+  DeclarePerStepDiscreteUpdateEvent(&SimulatorDrift::DiscreteVariableUpdate);
+  accumulated_drift_index_ =
+      this->DeclareDiscreteState(VectorXd::Zero(tree_.get_num_positions()));
+
+  time_idx_ = this->DeclareDiscreteState(1);
+  this->DeclareVectorOutputPort(OutputVector<double>(tree_.get_num_positions(),
+                                                     tree_.get_num_velocities(),
+                                                     tree_.get_num_actuators()),
+                                &SimulatorDrift::CalcAdjustedState);
+}
+
+drake::systems::EventStatus SimulatorDrift::DiscreteVariableUpdate(
+    const drake::systems::Context<double>& context,
+    drake::systems::DiscreteValues<double>* discrete_state) const {
+  const OutputVector<double>* state =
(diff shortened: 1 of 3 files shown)


examples/Cassie/simulator_drift.cc
```diff
@@ -0,0 +1,64 @@
...
+
+using dairlib::systems::OutputVector;
+using Eigen::VectorXd;
+
+SimulatorDrift::SimulatorDrift(const RigidBodyTree<double>& tree,
+                               const Eigen::VectorXd& drift_mean,
+                               const Eigen::MatrixXd& drift_cov)
+    : tree_(tree), drift_mean_(drift_mean), drift_cov_(drift_cov) {
+  DRAKE_ASSERT(drift_mean_.size() == drift_cov_.cols());
+  state_port_ = this
```
mposa: Why `->DeclareVectorInputPort(OutputVector<double>(` on a new line?

examples/Cassie/simulator_drift.cc
```diff
@@ -0,0 +1,64 @@
...
+}
+
+void SimulatorDrift::CalcAdjustedState(
+    const drake::systems::Context<double>& context,
+    dairlib::systems::OutputVector<double>* output) const {
+  const OutputVector<double>* robotOutput =
+      (OutputVector<double>*)this->EvalVectorInput(context, state_port_);
+  auto accumulated_drift =
+      context.get_discrete_state(accumulated_drift_index_).get_value();
+  VectorXd q = robotOutput->GetPositions();
```
mposa: Nitpicky (does not require changing).

This looks like a fairly inefficient way to do this, since it will allocate multiple VectorXd's. You should be able to just set the `output` directly from the `robotOutput`, and then modify positions.

examples/Cassie/simulator_drift.h
```diff
@@ -0,0 +1,35 @@
...
+
+#include <memory>
+#include <string>
+#include <vector>
+
+#include "drake/multibody/rigid_body_tree.h"
+#include "drake/systems/framework/leaf_system.h"
+#include "systems/framework/output_vector.h"
+
+class SimulatorDrift : public drake::systems::LeafSystem<double> {
```
mposa: Some documentation needed. This looks like it adds the output of a random walk to the position vector, correct? This should be stated, and along with requirements on the size of the mean and covariance.

examples/Cassie/simulator_drift.cc
```diff
@@ -0,0 +1,64 @@
...
+      (OutputVector<double>*)this->EvalVectorInput(context, state_port_);
+  auto prev_time_stamp =
+      discrete_state->get_mutable_vector(time_idx_).get_mutable_value();
+
+  if (state->get_timestamp() > prev_time_stamp(0)) {
+    auto accumulated_drift =
+        discrete_state->get_mutable_vector(accumulated_drift_index_)
+            .get_mutable_value();
+    VectorXd v = VectorXd::Random(drift_mean_.size());
+    accumulated_drift << accumulated_drift + (drift_mean_ + drift_cov_ * v);
```
mposa: Since it's not documented, I'm guessing a little, but shouldn't this be intended to emulate the integration of a stochastic process (random walk)? If so, the delta drift should depend on the length of the time between updates (or it should be set to update periodically, not perstep).

examples/Cassie/simulator_drift.cc
```diff
@@ -0,0 +1,59 @@
...
+
+void SimulatorDrift::CalcAdjustedState(
+    const drake::systems::Context<double>& context,
+    dairlib::systems::OutputVector<double>* output) const {
+  const OutputVector<double>* robotOutput =
+      (OutputVector<double>*)this->EvalVectorInput(context, state_port_);
+  auto accumulated_drift =
+      context.get_discrete_state(accumulated_drift_index_).get_value();
+
+  output->GetMutablePositions() << robotOutput->GetPositions() +
```
mposa: This does not set the velocities. Shouldn't it?

yangwill: Can the velocities accumulate error? I thought velocity estimates were more accurate because they're an instantaneous measurement.

I was also envisioning that any drift would be limited to positions that are wrt to the world frame (x, y, z, roll, pitch, yaw). Because those are the ones that we can't correct with the odometry sensors on Cassie.

mposa: Unless I missed something, I don't think the new version of the code sets the output velocities to _anything_.

yangwill: Yeah you're right, I misinterpreted what you said.

mposa:
Good to merge once the conflicts are resolved.

PR #118, opened 2019-11-25
Osc local walking

yminchen:
Main changes:
* Separated the high-level controller (which outputs desired yaw velocity and desired horizontal velocity) from the low-level walking controller, since the high-level controller will eventually be replaced by human command. See the diagram for the walking controller in `examples/Cassie/osc_walk`
https://github.com/DAIRLab/dairlib/blob/b317b86d0598723602265d585ad7f875fb018944/examples/Cassie/osc_walk/README.md 
* Added `SimulatorDrift` to the script `run_osc_walking_controller` to test that the controller is robust against the  drift in floating-base state

Minor changes:
* Fixed bugs in `SimulatorDrift`. It didn't copy over all the values of robot's output such as the timestamp.

The walking controller is sensitive to the swing foot height at the moment when the finite state machine switches to a new state. We can probably add a double support state to make it more robust. Posted Issue in #120


yminchen:
Just to make it clear, this PR is ready for review. @mposa

mposa:
I'll take a look at this, but would be good for @yangwill to weigh in, if possible.

examples/Cassie/osc_walk/heading_traj_generator.cc
```diff
@@ -0,0 +1,101 @@
...
+namespace dairlib {
+namespace cassie {
+namespace osc_walk {
+
+HeadingTrajGenerator::HeadingTrajGenerator(const RigidBodyTree<double>& tree,
+                                           int pelvis_idx)
+    : tree_(tree), pelvis_idx_(pelvis_idx) {
+  // Input/Output Setup
+  state_port_ = this
+                    ->DeclareVectorInputPort(OutputVector<double>(
```
yangwill: Minor comment: Drake's autoformat seems to have this spacing issue. I'll take a look to see if I can modify the .clang_format file to fix this.

examples/Cassie/osc_walk/heading_traj_generator.cc
```diff
@@ -0,0 +1,101 @@
...
+  // Get approximated heading angle of pelvis
+  Vector3d pelvis_heading_vec =
+      tree_.CalcBodyPoseInWorldFrame(cache, tree_.get_body(pelvis_idx_))
+          .linear()
+          .col(0);
+  double approx_pelvis_yaw =
+      atan2(pelvis_heading_vec(1), pelvis_heading_vec(0));
+
+  // Construct the PiecewisePolynomial.
+  const double one_sec = 1;
```
yangwill: Can you rename one_sec and make it a const member variable to make it easier to see how we're constructing the approximate trajectory?

examples/Cassie/osc_walk/heading_traj_generator.cc
```diff
@@ -0,0 +1,101 @@
...
+  double approx_pelvis_in_one_second =
+      approx_pelvis_yaw + des_yaw_vel(0) * one_sec;
+  Eigen::Vector4d pelvis_rotation_0(q(3), q(4), q(5), q(6));
```
yangwill: It's unclear what indices 3-6 represent, add a brief comment about what the indices represent whenever they're hardcoded.

examples/Cassie/osc_walk/high_level_command.cc
```diff
@@ -0,0 +1,207 @@
...
+          0;
+    }
+  }
+
+  return EventStatus::Succeeded();
+}
+
+void HighLevelCommand::CopyHeadingAngle(const Context<double>& context,
+                                    BasicVector<double>* output) const {
+  double desried_heading_pos =
```
yangwill: small typo

examples/Cassie/osc_walk/heading_traj_generator.cc
yminchen: I removed the variable `one_sec`, and added documentation to explain to algorithm

examples/Cassie/osc_walk/high_level_command.cc
yminchen: Fixed

examples/Cassie/osc_walk/high_level_command.cc
```diff
@@ -0,0 +1,207 @@
...
+    double desired_yaw =
+        atan2(global_com_pos_to_target_pos(1), global_com_pos_to_target_pos(0));
+
+    // Get current yaw velocity
+    double yaw_vel = v(2);
+
+    // PD position control
+    double des_yaw_vel =
+        kp_yaw_ * (desired_yaw - approx_pelvis_yaw) + kd_yaw_ * (-yaw_vel);
+    des_yaw_vel = std::min(vel_max_yaw_, std::max(vel_min_yaw_, des_yaw_vel));
```
yangwill: Minor comment - when we switch over to c++17, we should use [std::clamp](https://en.cppreference.com/w/cpp/algorithm/clamp) instead of these min(max) statements

examples/Cassie/osc_walk/high_level_command.cc
```diff
@@ -0,0 +1,207 @@
...
+
+    // Weigh between the desired yaw vel and 0 vel
+    double weight = 1 / (1 + exp(-params_of_no_turning_(0) *
+        (global_com_pos_to_target_pos.norm() -
+            params_of_no_turning_(1))));
+    double desired_filtered_yaw_vel =
+        (1 - weight) * 0 + weight * des_yaw_vel;
```
yangwill: Why are we multiplying by 0 here?

examples/Cassie/osc_walk/heading_traj_generator.cc
```diff
@@ -0,0 +1,105 @@
...
+
+  const std::vector<double> breaks = {context.get_time(),
+                                      context.get_time() + 0.1};
+  std::vector<MatrixXd> knots(breaks.size(), MatrixXd::Zero(4, 1));
+  knots[0] = pelvis_rotation_i;
+  knots[1] = pelvis_rotation_f;
+  const auto pp = PiecewisePolynomial<double>::FirstOrderHold(breaks, knots);
+
+  // Assign traj
+  PiecewisePolynomial<double>* casted_traj =
```
yangwill: I could be wrong, but I don't think you need to cast the trajectory before assigning

systems/controllers/cp_traj_gen.cc
```diff
@@ -266,8 +270,8 @@ PiecewisePolynomial<double> CPTrajGenerator::createSplineForSwingFoot(
   Y[2](1, 0) = CP(1);
   // z
   Y[0](2, 0) = init_swing_foot_pos(2);
-  Y[1](2, 0) = mid_foot_height_;
-  Y[2](2, 0) = desired_final_foot_height_;
+  Y[1](2, 0) = mid_foot_height_ + stance_foot_height(0);
```
yangwill: Can you add a comment about why we're adding the z-component of the stance foot to the spline?

examples/Cassie/osc_walk/high_level_command.cc
yminchen: To explicitly show that it is a convex combination of 0 and des_yaw_vel

examples/Cassie/osc_walk/high_level_command.cc
yminchen: Sounds good!

examples/Cassie/osc_walk/heading_traj_generator.cc
yminchen: If we don't cast it, we get an error
```
constexpr drake::trajectories::Trajectory<T>& drake::trajectories::Trajectory<T>::operator=(const drake::trajectories::Trajectory<T>&) [with T = double]' is protected within this context
```
because they are different types.

systems/controllers/cp_traj_gen.cc
yminchen: done

examples/Cassie/osc_walk/high_level_command.cc
```diff
@@ -0,0 +1,207 @@
...
+  auto delta_CP_3D_global =
+      context.get_discrete_state(des_horizontal_vel_idx_).get_value();
+
+  // Assign
+  output->get_mutable_value() = delta_CP_3D_global;
+}
+
+}  // namespace osc_walk
+}  // namespace cassie
+}  // namespace dairlib
```
mposa: Missing new line at end of file

examples/Cassie/osc_walk/heading_traj_generator.cc
yminchen: I removed `const` from `pp`, but I still got the same error message. Code:
```
auto pp = PiecewisePolynomial<double>::FirstOrderHold(breaks, knots);
*traj = pp;
```

examples/Cassie/osc_walk/high_level_command.cc
yminchen: Added

examples/Cassie/osc_walk/high_level_command.cc
yminchen: `weight` is a variable. I added a line of comment to explain that this is a convex combination of 0 and `des_yaw_vel`

PR #119, opened 2019-12-03
Explicitly connecting output ports in dispatcher_in

yangwill:
Due to the input supervisor now outputting its status (now has 2 output ports), we must explicitly connect its output ports.

Note. The status port of the input supervisor is not connected to anything.


mposa:
Good catch. Should we add output port getters to the input supervisor, rather than the rather opaque 0 index?

PR #121, opened 2019-12-06
Sync channel switch with controller

yminchen:
Built on PR #114. 

Since the `TimeBasedFiniteStateMachine` uses the global time, we can not switch to the controller (e.g. the current osc walking controller) that uses the time-based finite state machine at arbitrary time. Instead, it has to happens at the same time as the finite state machine switches to a new state. 

Changes:
* Modify `run_controller_switch.cc` to control when to publish the switch lcm message.
* Delete `ControllerChannelSender` since we can just implement the functionality outside leafsystem
* Update LcmDrivenLoop so that we can set simulation duration, and also can run diagram without lcm message parser (one use case is that we only need time from lcm message)
* Fix a bug (publishing the same lcm message multiple times) in LcmDrivenLoop.


mposa:
Can you resolve conflicts/merge with master first? This contains a lot of #114 in its diff.

yminchen:
> Can you resolve conflicts/merge with master first? This contains a lot of #114 in its diff.

Done. I thought git could have merged the branches automatically, because this branch is just #114 plus more changes.

yminchen:
It's probably better to not implement ControllerChannelSender as a leafsystem based on the discussion with @mposa . I will make the changes and then request review.

yminchen:
@mposa This PR is ready for review

systems/framework/lcm_driven_loop.h
```diff
@@ -206,7 +227,13 @@ class LcmDrivenLoop {
 
         simulator_->AdvanceTo(time);
         // Force-publish via the diagram
-        diagram_ptr_->Publish(diagram_context);
+        /// Notice that diagram.Publish() is for dispatching the publish of
+        /// TriggerType::kForced type. In LcmPublisherSystem, both periodic and
+        /// per-step publishes are also forced publish.
+        /// https://github.com/RobotLocomotion/drake/blob/03fe7e4/systems/lcm/lcm_publisher_system.h#L54-L56
```
mposa: This is not an entirely correct interpretation of LcmPublisherSystem--I fixed this issue in Drake some time ago. See the correct constructor https://github.com/RobotLocomotion/drake/blob/master/systems/lcm/lcm_publisher_system.h#L111 which allows you to specify exactly which publish types to use (e.g. forced only)

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
 
-DEFINE_double(end_time, std::numeric_limits<double>::infinity(),
-              "End time of simulation");
-DEFINE_double(publish_rate, 1000, "Publishing frequency (Hz)");
+DEFINE_string(channel_x, "CASSIE_STATE",
+              "The name of the channel which receives state");
 DEFINE_string(controller_channel, "PD_CONTROLLER",
               "The name of the lcm channel that dispatcher_in listens to");
+DEFINE_int32(n_fsm_period, -1,
+             "the number of state-switch after the start of the"
```
mposa: I'm not sure what "number of state-switch" means. I think there's a grammatical error here.

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
 
-/// This diagram publishes a string which tells dispatcher_robot_in which
-/// channel to listen to.
+/// The program is extended so that it could be used with
+/// `TimeBasedFiniteStateMachine`. More specifically, the string assigment
+/// starts when `TimeBasedFiniteStateMachine` switches to a new state (because
+/// we don't want to switch to a new controller when the new controller is in a
+/// middle of a discrete state). This requires that the users provide two (or
+/// three) more arguments to the constructors besides `controller_channel`:
+///   @param n_fsm_period, the number of state-switch after the start of the
```
mposa: Ditto above.

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
 #include <string>
-
 #include <gflags/gflags.h>
+#include "dairlib/lcmt_controller_switch.hpp"
+#include "dairlib/lcmt_robot_output.hpp"
+#include "systems/framework/lcm_driven_loop.h"
```
mposa: No need for lcm_driven_loop (and maybe other dependencies)

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
+  // Ensure that if (n_fsm_period >= 0), then (period > 0).
+  DRAKE_DEMAND((FLAGS_n_fsm_period < 0) || (FLAGS_fsm_period > 0));
+  // offset has to be positive
+  DRAKE_DEMAND(FLAGS_fsm_offset >= 0);
+
+  // Parameters
+  using PublishMessageType = dairlib::lcmt_controller_switch;
+  using TriggerMessageType = dairlib::lcmt_robot_output;
+  std::string switch_channel = "INPUT_SWITCH";
+  int n_publishes = 10;
```
mposa: This should probably also be a command line option, with a description.

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
 using drake::systems::lcm::LcmPublisherSystem;
+using drake::systems::lcm::TriggerTypeSet;
 
-DEFINE_double(end_time, std::numeric_limits<double>::infinity(),
-              "End time of simulation");
-DEFINE_double(publish_rate, 1000, "Publishing frequency (Hz)");
+DEFINE_string(channel_x, "CASSIE_STATE",
+              "The name of the channel which receives state");
 DEFINE_string(controller_channel, "PD_CONTROLLER",
               "The name of the lcm channel that dispatcher_in listens to");
```
mposa: This comment isn't very clear. This is the new channel, after the switch.

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
 
+  // Ensure that if (n_fsm_period >= 0), then (period > 0).
+  DRAKE_DEMAND((FLAGS_n_fsm_period < 0) || (FLAGS_fsm_period > 0));
+  // offset has to be positive
+  DRAKE_DEMAND(FLAGS_fsm_offset >= 0);
+
+  // Parameters
+  using PublishMessageType = dairlib::lcmt_controller_switch;
+  using TriggerMessageType = dairlib::lcmt_robot_output;
+  std::string switch_channel = "INPUT_SWITCH";
```
mposa: Given that all channels are settable by input argument, this probably should be as well.

examples/Cassie/run_controller_switch.cc
```diff
@@ -1,55 +1,136 @@
...
 int do_main(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
 
+  // Ensure that if (n_fsm_period >= 0), then (period > 0).
+  DRAKE_DEMAND((FLAGS_n_fsm_period < 0) || (FLAGS_fsm_period > 0));
+  // offset has to be positive
+  DRAKE_DEMAND(FLAGS_fsm_offset >= 0);
+
+  // Parameters
+  using PublishMessageType = dairlib::lcmt_controller_switch;
```
mposa: These `using` definitions, mid-function, hurt readability. Are they really necessary?

systems/framework/lcm_driven_loop.h
yminchen: It's a little bit ambiguous here. What I meant was we cannot run both `AdvanceTo()` and `Publish()` in the case of periodic or per-step publisher. 

After reading the code for the second time, I should probably add a flag which allows users to run `Publish()`

examples/Cassie/run_controller_switch.cc
yminchen: Finished cleaning up the include's

examples/Cassie/run_controller_switch.cc
yminchen: Done

examples/Cassie/run_controller_switch.cc
yminchen: Done

examples/Cassie/run_controller_switch.cc
yminchen: Done

examples/Cassie/run_controller_switch.cc
yminchen: It's not necessary. I wrote it like because I was thinking of generalizing the code.

systems/framework/lcm_driven_loop.h
```diff
@@ -114,26 +114,42 @@ class LcmDrivenLoop {
...
+      : LcmDrivenLoop(drake_lcm, std::move(diagram), nullptr,
+                      std::vector<std::string>(1, input_channel), input_channel,
+                      ""){};
+
+  // Start simulating the diagram
+  void Simulate() { Simulate(std::numeric_limits<double>::infinity(), true); }
+
   // Start simulating the diagram
-  void Simulate(double end_time = std::numeric_limits<double>::infinity()) {
+  /// @param is_end_time_or_duration, true if `time_in` is the end time of the
```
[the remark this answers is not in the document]
yminchen: I made the old `Simulate()` function private and created public functions `SimulateTo` and `SimulateDuration`

examples/Cassie/run_controller_switch.cc
yminchen: The description was not updated after I changed the variable name. That's why it doesn't make sense. `n_fsm_period` is the number of periods (of the finite state machine) before we start publishing the new channel name.

I also updated the documentation. Hope it's easier to understand now

yminchen:
To make it clear, this branch is ready for another review.

systems/framework/lcm_driven_loop.h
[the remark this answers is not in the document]
mposa: I feel like this approach is still awkward (it's the same code, just hidden now). Is there a reason you think the approach I suggested is worse?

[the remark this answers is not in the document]
yminchen: I agree that the approach looks awkward, and since we are not using it now, I think I'll just delete it.

The reason I created `delta T` aside from the original `final T` was because the old version of`run_controller_switch.cc` used `LcmDrivenLoop` and we want to turn off `run_controller_switch.cc` after a duration `delta T`.

As for the API, the design arises from minimizing duplicated code. Since the difference between those two modes is just one line code, I decided to leave most of the old `Simulate()` method untouched, and created two functions that call it.

PR #123, opened 2019-12-13
Python scripts for processing ROS odometry data into a CSV file with experiments annotated.

mshalm:
Two scripts for processing motion data captured as ROS odom messages stored in a rosbag:

`apriltag_csv.py` is a python script that converts the .bag format data into CSV time series data directly without post-processing.
`process_dynamics.py` is a python script that converts a .bag into full-state data for block tossing experiments by
1. Running `apriltag_csv.py`;
2. Using scipy interpolation tools to differentiate the raw pose data;
3. Using scipy signal tools to low-pass filter the resulting velocity data;
4. Detecting + labeling the starting & ending timestamps for the various tosses contained in the rosbag.

Both scripts were tested on real tossing data.

### DIFF
--- a/tools/apriltag/process_dynamics.py
+++ b/tools/apriltag/process_dynamics.py
@@ -1,0 +1,203 @@
+
+import csv
+import os
+import pdb
+import sys
+import time
+
+import numpy as np
+
+from scipy import signal
+from scipy.spatial.transform import Rotation, RotationSpline
+from scipy.interpolate import CubicSpline
+
+import matplotlib.pyplot as plt
+
+if len(sys.argv) < 3:
+    print("usage: python[3] process_dynamics.py [INBAG] [OUTCSV]")
+    sys.exit()
+
+# static parameters for toss detection
+TOSS_BALL_TOLERANCE = 1e-3
+TOSS_BALL_H = 10
+
+# far edge of board in board coordinates (y-axis)
+BOARD_EDGE_Y = 0.6
+
+# true if datapoint is over the board
+OFFBOARD_CONDITION = lambda data, i: data[i,1] > BOARD_EDGE_Y
+
+# false if datapoint is over the board
+ONBOARD_CONDITION = lambda data, i: not OFFBOARD_CONDITION(data, i)
+
+# true if block moves significantly over horizon H
+MOVING_CONDITION = lambda data, i: minBallOverHorizon(data[i:,:], TOSS_BALL_H) > TOSS_BALL_TOLERANCE
+
+# false if block moves significantly over horizon H
+STOPPED_CONDITION = lambda data, i: not MOVING_CONDITION(data, i)
+
+# returns minium ||\delta x|\ over a horizon of H in data
+def minBallOverHorizon(data, H):
+    diff_data = data[1:H,:] - data[0:(H-1),:]
+    return np.max(np.linalg.norm(diff_data,axis=1))
+
+# find first index of data for which condition() returns true
+def get_first_true(data,condition,H):
+    i = 0
+    while i<(data.shape[0]-(H-1)):
+        if condition(data, i):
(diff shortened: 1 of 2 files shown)


tools/apriltag/process_dynamics.py
```diff
@@ -0,0 +1,203 @@
...
+# true if block moves significantly over horizon H
+MOVING_CONDITION = lambda data, i: minBallOverHorizon(data[i:,:], TOSS_BALL_H) > TOSS_BALL_TOLERANCE
+
+# false if block moves significantly over horizon H
+STOPPED_CONDITION = lambda data, i: not MOVING_CONDITION(data, i)
+
+# returns minium ||\delta x|\ over a horizon of H in data
+def minBallOverHorizon(data, H):
+    diff_data = data[1:H,:] - data[0:(H-1),:]
+    return np.max(np.linalg.norm(diff_data,axis=1))
```
spfrommer: Slightly confused why this returns the minimum Dx over a span of data. Shouldn't the return line read np.min?

spfrommer: Alternatively, I feel like you might actually want the max movement based on the MOVING_CONDITION above, in which case maybe the function name should be adjusted

tools/apriltag/process_dynamics.py
```diff
@@ -0,0 +1,203 @@
...
+
+# find first index of data for which condition() returns true
+def get_first_true(data,condition,H):
+    i = 0
+    while i<(data.shape[0]-(H-1)):
+        if condition(data, i):
+            return i
+        i = i+1
```
spfrommer: More pythonic would be:
```
for i in range(0, data.shape[0] - (H-1))
```

tools/apriltag/process_dynamics.py
```diff
@@ -0,0 +1,203 @@
...
+        return (-1,0)
+    end_time = stopped_time + TOSS_BALL_H
+    return (start_time,end_time)
+
+# permutation of get_first_experiment to start at index start
+def get_first_experiment_after(start, data):
+    (s,e) = get_first_experiment(data[start:,:])
+    s = s + start
+    e = e + start
+    return (s,e)
```
spfrommer: Could probably just combine these into return(s+start, e+start)

tools/apriltag/process_dynamics.py
```diff
@@ -0,0 +1,203 @@
...
+print('python2 apriltag_csv.py ' + inbag + ' ' + csvfn)
+os.system('python2 apriltag_csv.py ' + inbag + ' ' + csvfn)
+outcsv = sys.argv[2]
+
+# get number of datapoints
+num_t = 0
+with open(csvfn,'r') as csvfile:
+    plots = csv.reader(csvfile, delimiter=',')
+    for row in plots:
+        num_t = num_t + 1
```
spfrommer:
```
num_t += len(plots)
```

tools/apriltag/process_dynamics.py
```diff
@@ -0,0 +1,203 @@
...
+with open(csvfn,'r') as csvfile:
+    plots = csv.reader(csvfile, delimiter=',')
+    for row in plots:
+        row = np.asarray([float(e) for e in row])
+        t[i] = row[0]
+        q_t[:,i] = row[1:5]
+        p_t[:,i] = row[5:8]
+        bq_t[:,i] = row[8:12]
+        bp_t[:,i] = row[12:15]
+        i = i + 1
```
spfrommer: Probably could also use
```
for i, row in enumerate(plots)
```

tools/apriltag/process_dynamics.py
mshalm: renamed to maxDeltaOverHorizon

Issue #115, opened 2019-11-18
Benchmarks of Dircon performance


yminchen:
While we are modifying Dircon, we benchmark the performance here for future reference.

CPU: Intel(R) Core(TM) i7-8700K CPU @ 3.70GHz
Program: run_dircon_squatting.cc

Commit: ea86f212ed205ceb7a5809ea10f726423d284948
Snopt solve time: 125.758 seconds

yminchen:
There are some benchmark tests done in #169. We can see that a single change to the traj opt problem doesn't necessarily improve the solve time.

yminchen:
I compared ipopt with snopt in the walking example (https://github.com/DAIRLab/dairlib/blob/5f5205cbfc9466531bcefc10ad524222fd0998fb/examples/Cassie/run_dircon_walking.cc) in my local branch. 

Ipopt setting: 
```
    // Ipopt settings adapted from CaSaDi and FROST
    auto id = drake::solvers::IpoptSolver::id();
    trajopt.SetSolverOption(id, "tol", tol);
    trajopt.SetSolverOption(id, "dual_inf_tol", tol);
    trajopt.SetSolverOption(id, "constr_viol_tol", tol);
    trajopt.SetSolverOption(id, "compl_inf_tol", tol);
    trajopt.SetSolverOption(id, "max_iter", max_iter);
    trajopt.SetSolverOption(id, "nlp_lower_bound_inf", -1e6);
    trajopt.SetSolverOption(id, "nlp_upper_bound_inf", 1e6);
    trajopt.SetSolverOption(id, "print_timing_statistics", "yes");
    trajopt.SetSolverOption(id, "print_level", 5);

    // Set to ignore overall tolerance/dual infeasibility, but terminate when
    // primal feasible and objective fails to increase over 5 iterations.
    trajopt.SetSolverOption(id, "acceptable_compl_inf_tol", tol);
    trajopt.SetSolverOption(id, "acceptable_constr_viol_tol", tol);
    trajopt.SetSolverOption(id, "acceptable_obj_change_tol", 1e-3);
    trajopt.SetSolverOption(id, "acceptable_tol", 1e2);
    trajopt.SetSolverOption(id, "acceptable_iter", 5);
```
Snopt setting:
```
    auto id = drake::solvers::SnoptSolver::id();
    // trajopt.SetSolverOption(id, "Print file", "../snopt.out");
    trajopt.SetSolverOption(id, "Major iterations limit", max_iter);
[...]

PR #125, opened 2019-12-16
Scale decision variable

yminchen:
(This PR is built upon  #124)
Scale decision variable to improve dircon. 

Scaling decision variable properly helps with solution convergence, as stated in SNOPT documentation. Also see #115 for performance benchmarks.

Changes: 
* A small update in DirconAbstractConstraint (kind of a bug)
* Add setters for variable scaling in HybridDircon
* Apply variable scaling in the squatting example


yminchen:
The failure of the tests could probably be caused by CI's bazel version?

yminchen:
To make it clear, this PR is ready for review.

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -213,8 +217,19 @@ void GetInitFixedPointGuess(const Vector3d& pelvis_position,
       -1.762, 68.42, 0.652, 1.744;
   mp->SetInitialGuessForAllVariables(init_guess);
 
-  // mp->SetSolverOption(drake::solvers::SnoptSolver::id(),
-  //                     "Print file", "../snopt.out");
+  // Test scaling
```
mposa: Lets not commit this dead test code to master.

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -97,11 +97,15 @@ using dairlib::multibody::GetBodyIndexFromName;
...
               "directory to save/read data");
-DEFINE_bool(store_data, false, "To store soluation or not");
+DEFINE_bool(store_data, false, "To store solution or not");
 DEFINE_int32(max_iter, 100000, "Iteration limit");
 DEFINE_double(duration, 0.4, "Duration of the single support phase (s)");
 DEFINE_double(tol, 1e-4, "Tolerance for constraint violation and dual gap");
 
+// Parameters which enable dircon-improving features
+DEFINE_bool(is_scale_constraint, true, "Scale the nonlinear constraint values");
+DEFINE_bool(is_scale_variable, false, "Scale the decision variable");
```
mposa: Is false the right default?

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -551,6 +581,48 @@ void DoMain(double duration, int max_iter, string data_directory,
...
+    trajopt->ScaleForceVariables(10, 0, 0, 1);
+    trajopt->ScaleForceVariables(1000, 0, 2, 2); //1000
+    trajopt->ScaleForceVariables(10, 0, 3, 4);
+    trajopt->ScaleForceVariables(1000, 0, 5, 5);
+    trajopt->ScaleForceVariables(10, 0, 6, 7);
+    trajopt->ScaleForceVariables(1000, 0, 8, 8);
+    trajopt->ScaleForceVariables(10, 0, 9, 10);
+    trajopt->ScaleForceVariables(1000, 0, 11, 11);
+    trajopt->ScaleForceVariables(600, 0, 12, 13);
+//    trajopt->ScaleQuaternionSlackVariables(0.5);
```
mposa: Ditto on the commented out code below.

systems/trajectory_optimization/hybrid_dircon.h
```diff
@@ -134,7 +137,27 @@ class HybridDircon
...
+  void ScaleImpulseVariables(double scale, int mode, int idx_start,
+                             int idx_end);
+  void ScaleQuaternionSlackVariables(double scale);
+  void ScaleKinConstraintSlackVariables(double scale, int mode, int idx_start,
+                                        int idx_end);
+
+  /**
+   * Check whether the scaling of decision variable @p var is set or not .
+   * @param var the decision variable
+   * @return true if all corresponding scaling factors are not set yet.
```
mposa: Returns true if the scaling is not set? This seems backwards with the name. Also, it says here "all corresponding scaling factors". What does "all" mean? This seems to be a function only for one varaible.

systems/trajectory_optimization/hybrid_dircon.h
```diff
@@ -134,7 +137,27 @@ class HybridDircon
 
   using drake::systems::trajectory_optimization::MultipleShooting::N;
   using drake::systems::trajectory_optimization::MultipleShooting::
-        SubstitutePlaceholderVariables;
+      SubstitutePlaceholderVariables;
+
+  void ScaleTimeVariables(double scale);
+  void ScaleStateVariables(double scale, int idx_start, int idx_end);
```
mposa: I can't say I'm a fan of this API. Two reasons:

1. The Drake API is `setScaling(variable, scale)`, which I think makes sense. Here, you've reversed that ordering.
2. "start, end" is a very old-fashioned approach. I suppose it's OK, but I think the more basic `ScaleXXVariable(idx, scale)` should also be included (for a single variable) as an option.

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -418,6 +419,129 @@ void HybridDircon<T>::SetInitialForceTrajectory(
   SetInitialGuess(collocation_slack_vars_[mode], guess_collocation_slack);
 }
 
+template <typename T>
+void HybridDircon<T>::ScaleTimeVariables(double scale) {
+  for (int i = 0; i < h_vars().size(); i++) {
+    DRAKE_DEMAND(IsVariableScalingUnset(h_vars()(i)));
```
mposa: What's wrong with re-writing the scale factor? Why is this DEMAND necessary?

examples/Cassie/run_dircon_squatting.cc
yminchen: I set it to false, because it runs slower with variable scaling in this example. It is possible that the initial guess in the example is very good, so the variable scaling doesn't help much. (it's hard to find the optimal scaling factors that improve the performance)

examples/Cassie/run_dircon_squatting.cc
yminchen: Done

examples/Cassie/run_dircon_squatting.cc
yminchen: Done

systems/trajectory_optimization/hybrid_dircon.h
yminchen: You are correct that this only check one variable. I missed the documentation when modifying the code. 

This function returns true if the variable is not set. The documentation is cohesive with the code

systems/trajectory_optimization/hybrid_dircon.h
yminchen: To scale a variable, we could use `SetVariableScaling` from Drake. I am curious what you think is a less old-fashioned approach. Maybe ScaleXXVariable(idx_list, scale)?

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: I used DEMAND to avoid unintentional rewriting. I guess I can remove it so that it's not so restrictive

systems/trajectory_optimization/hybrid_dircon.h
mposa: I see the functions you've created here a syntactic sugar on top of Drake's `SetVariableScaling` (ways to auto-extract all the state/time/input/etc. variables).

I think what you mentioned `ScaleXXVariable(idx, scale)` makes sense, yes (mentioned in my earlier comment). If you think it's important to be able to set ranges of variables in one call, then I suppose `start,end` is OK. The other alternative would be something like `ScaleXXVariables(vector<int> indices, scale)`, but that's not a very easy API.

systems/trajectory_optimization/hybrid_dircon.h
mposa: `IsUnset` seems to add an unnecessary negative. Why `IsUnset` vs . `IsSet`? Also, is this method even needed anymore?

systems/trajectory_optimization/hybrid_dircon.cc
mposa: I think we can trust users to rewrite only if they mean it.

systems/trajectory_optimization/hybrid_dircon.h
yminchen: No, I just deleted them, since we are not using it now

systems/trajectory_optimization/hybrid_dircon.h
yminchen: I changed the API to `ScaleXXVariable(idx_list, scale)`. I think it's not hard to use

systems/trajectory_optimization/hybrid_dircon.cc
yminchen: I removed `DRAKE_DEMAND`. It's easy for users to check if a variable scaling has been set

yminchen:
@mposa I made the changes according to your comments. It's ready for another review

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -578,23 +580,31 @@ void DoMain(double duration, int max_iter, string data_directory,
...
-    trajopt->ScaleForceVariables(1000, 0, 5, 5);
-    trajopt->ScaleForceVariables(10, 0, 6, 7);
-    trajopt->ScaleForceVariables(1000, 0, 8, 8);
-    trajopt->ScaleForceVariables(10, 0, 9, 10);
-    trajopt->ScaleForceVariables(1000, 0, 11, 11);
-    trajopt->ScaleForceVariables(600, 0, 12, 13);
+    trajopt->ScaleForceVariables(0, {0, 1}, 10);
+    trajopt->ScaleForceVariables(0, {2, 2}, 1000);  // 1000
+    trajopt->ScaleForceVariables(0, {3, 4}, 10);
+    trajopt->ScaleForceVariables(0, {5, 5}, 1000);
```
mposa: Some of these only scale a single variable, like `{5, 5}` should either be
`ScaleForceVariables(0, {5}, 1000)`; 
or (probably better)
`ScaleForceVariable(0, 5, 1000);`

yminchen: I intentionally left it like that cause it aligned beautifully... but I guess I shouldn't let OCD come into play here!

Just updated.

PR #126, opened 2019-12-27
Add double support phase for Cassie walking controller

yminchen:
Changes:
* Extend `TimeBasedFiniteStateMachine` to more than 2 states
* Add double support phase to Cassie's walking controller
* Fix a bug in lipm trajectory generation.
* Tune controller gains


yminchen:
@mposa This branch is ready for review

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -55,6 +59,8 @@ DEFINE_string(channel_u, "CASSIE_INPUT",
 int DoMain(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
 
+  bool is_two_phase = false;
```
yangwill: This should probably be a flag

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -138,21 +144,66 @@ int DoMain(int argc, char* argv[]) {
...
   int left_stance_state = 0;
   int right_stance_state = 1;
-  double duration_per_state = 0.35;
-  double time_shift = 0;
+  int double_support_state = 2;
+  double left_support_duration = 0.35;
+  double right_support_duration = 0.35;
+  double double_support_duration = 0.02;
+  std::vector<int> fsm_states;
+  std::vector<double> state_druations;
```
yangwill: small typo

examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -138,21 +144,66 @@ int DoMain(int argc, char* argv[]) {
...
   auto fsm = builder.AddSystem<systems::TimeBasedFiniteStateMachine>(
-      tree_with_springs, duration_per_state, time_shift);
+      tree_with_springs, fsm_states, state_druations);
   builder.Connect(simulator_drift->get_output_port(0),
                   fsm->get_input_port_state());
 
   // Create CoM trajectory generator
   double desired_com_height = 0.89;
+  std::vector<int> unordered_fsm_states;
+  std::vector<double> unordered_state_druations;
```
yangwill: typo here also

systems/controllers/cp_traj_gen.cc
```diff
@@ -10,77 +10,76 @@ using std::cout;
...
-    right_foot_idx_(right_foot_idx),
-    pt_on_left_foot_(pt_on_left_foot),
-    pt_on_right_foot_(pt_on_right_foot),
-    pelvis_idx_(pelvis_idx),
-    add_extra_control_(add_extra_control),
-    is_feet_collision_avoid_(is_feet_collision_avoid),
-    is_using_predicted_com_(is_using_predicted_com),
-    cp_offset_(cp_offset),
-    center_line_offset_(center_line_offset) {
+CPTrajGenerator::CPTrajGenerator(
```
yangwill: This constructor is getting really big, is there a way to condense down the number of parameters? To start, we could probably remove some of the bools (is_feet_collision_avoid). Some parameters like mid_foot_height, max_CoM_to_CP_dist could be set as const static member variables instead.

yminchen: I think all of the parameters are essential and shouldn't be hard-coded, but I combine the right and left leg index/position and states of the finite state machine.

examples/Cassie/run_osc_walking_controller.cc
yminchen: done

examples/Cassie/run_osc_walking_controller.cc
yminchen: fixed

examples/Cassie/run_osc_walking_controller.cc
yminchen: fixed

systems/controllers/cp_traj_gen.cc
yminchen: After face-to-face discussion with @yangwill, we decided to leave it as is

PR #127, opened 2020-01-20
Add sparsity pattern

yminchen:
Built upon #125 (Commit af0fc3b)


yminchen:
Closed the pull request because the sparsity pattern constructor cannot be generalized to deal with constraints which contain if-statement and min/max functions.

PR #128, opened 2020-01-31
Updating master to work with newest version of Drake

yangwill:
As every computer has been updated to Ubuntu 18.04 and because Yu-ming and I (Will) have been working with newer versions of Drake, I think we should update master to build with the newest version of Drake.

Updating only involved a few changes:
- Update .bazelrc to use c++17
- Use std versions where applicable to avoid Drake deprecations
- Update drake-visualizer to use python3.6 instead of python2

### DIFF
--- a/examples/PlanarWalker/test_contact.cc
+++ b/examples/PlanarWalker/test_contact.cc
@@ -10,7 +10,6 @@
 #include "drake/multibody/kinematics_cache.h"
 #include "drake/math/autodiff.h"
 #include "drake/math/autodiff_gradient.h"
-#include "drake/systems/controllers/controlUtil.h"
 
 using Eigen::VectorXd;
 using Eigen::MatrixXd;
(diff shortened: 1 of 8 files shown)


examples/PlanarWalker/test_contact.cc
```diff
@@ -148,13 +147,13 @@ int do_main(int argc, char* argv[]) {
   tree.get()->surfaceTangents(n_world, d_world);
 
   Eigen::Matrix<double,3,2> d;
-  surfaceTangents(normal.col(0),d);
+  //  surfaceTangents(normal.col(0),d);
```
yminchen: Why is this line commented out?

yangwill: The original method is being deprecated. I didn't think it was worth replicating the method to test a rigidbodytree function.

PR #134, opened 2020-03-10
Implemented MBP simulator with stiff springs to simulate loop closures

yangwill:
The mbp_sim_dev branch got cluttered with commits. These are the only useful changes from that branch.

The timestepping simulator runs at real time with a time step of 8e-5 and doesn't seem to have any issues.

The continuous simulator runs at ~0.5 real time. It has stability issues around large impacts (ie dropping from a tall height > 1.3m) and also hasn't implemented joint limits so it can also blow up if the joints fall through each other.

### DIFF
--- a/examples/Cassie/multibody_sim.cc
+++ b/examples/Cassie/multibody_sim.cc
@@ -173,10 +173,15 @@ Eigen::VectorXd GetInitialState(const MultibodyPlant<double>& plant) {
   std::map<std::string, int> positions_map =
       multibody::makeNameToPositionsMap(plant);
 
-  VectorXd q_ik_guess = VectorXd::Zero(plant.num_positions());
+  for (auto pair : positions_map) {
+    std::cout << pair.first << ": " << pair.second << std::endl;
+  }
+
+  VectorXd q_ik_guess = VectorXd::Zero(n_q);
   Eigen::Vector4d quat(1, 0, 0, 0);
   q_ik_guess << quat.normalized(), 0.001, 0.001, 1.1, -0.01, 0.01, 0.0, 0.0,
-      1.15, 1.15, -1.35, -1.35, 1.0, 1.0, -M_PI / 2, -M_PI / 2;
+      1.15, 1.15, -1.35, -1.35, 1.0, 1.0, 0.0, 0.0, 0.0, -M_PI / 2, 0.0,
+      -M_PI / 2;
 
   double eps = 1e-3;
   Vector3d eps_vec = eps * VectorXd::Ones(3);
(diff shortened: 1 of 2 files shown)


examples/Cassie/multibody_sim.cc
```diff
@@ -146,20 +156,80 @@ int do_main(int argc, char* argv[]) {
...
   simulator.set_target_realtime_rate(FLAGS_target_realtime_rate);
   simulator.Initialize();
-  simulator.AdvanceTo(std::numeric_limits<double>::infinity());
+  simulator.AdvanceTo(FLAGS_end_time);
 
   return 0;
 }
 
-}  // namespace dairlib
+Eigen::VectorXd GetInitialState(const MultibodyPlant<double>& plant) {
```
mposa: Can you comment a little on the goal of this function, and some of the constraints below? Some specific questions:
- Does this function solve for an equilibrium configuration?
- How do the linear constraints below enforce the loop closure (which would seem nonlinear)?

yangwill: The goal of the function is to make it easier to set the initial configuration for the simulated robot (ie set toe/pelvis positions). I don't think the IK solver necessarily solves for a configuration that is in equilibrium but just from testing, any deviations from equilibrium get damped out quickly (the OSC standing controller reaches equilibrium in 0.2s).

mposa: What about the loop closure? Also, the toes don't seem to have to be exactly on the ground. Why is this?

yangwill: I'm confused about the loop closure - the loop closure constraints are still implemented in addCassieMultibody in cassie_utils .cc. This function only solves for the initial positions for the simulator and doesn't impose any constraints on the simulator.

The toes start slightly above the ground because the starting configuration isn't perfectly flush with the ground (as in there are 4 contacts). This should be changed, I'll add a change to the IK solver to constrain all 4 contact points to be on the ground.

mposa: You would probably want the initial state to not have an energy in the loop closure springs (e.g. the loops are properly initialized in a closed configuration). In RBT, this (and finding equlibria) was done via a custom Mathematical Program: https://github.com/DAIRLab/dairlib/blob/master/attic/multibody/multibody_solvers.cc

mposa: Going to approve, though it's probably a TODO to replicate the fixed point solver from RBT.

yangwill:
I'm actually going to wait for hongkai to add a distance constraint option in the IK solver, that way we can at least remove the energy in the loop closure.

yangwill:
This uses methods from a newer version of Drake, so we will have to update the commit number used by master before we can merge

mposa:
This PR should be on the one to update the commit #.

PR #142, opened 2020-04-12
Adding function to construct vector from maps

yangwill:
I've need to use this functionality several times and had to copy in each case. It builds from the functions that already exist in multibody_utils so I thought it would fit well there.

### DIFF
--- a/multibody/multibody_utils.cc
+++ b/multibody/multibody_utils.cc
@@ -233,25 +233,28 @@ map<string, int> makeNameToActuatorsMap(const MultibodyPlant<T>& plant) {
   return name_to_index_map;
 }
 
-vector<string> createStateAndActuatorNameVectorFromMap(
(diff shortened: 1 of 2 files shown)


yangwill:
Good point. I can, but because there aren't any requirements on the map, the user can just input an empty `std::map<std::string, int>` for any fields they don't want converted. I'll add that in the documentation.

yangwill:
Or would it make more sense to just have a single function that converts any std::map to a vector of strings? The occasions that I use this function are mainly to label the states of the trajectories I'm saving.

yminchen:
I prefer the original implementation, since this function only converts a `map<string, int>` of which the `int` is non-repeating and "continuous" (no missing integer between any two)

yangwill:
I changed the function to treat the state and inputs separately - I agree that it makes more sense this way.

PR #143, opened 2020-04-22
Add python bindings for FindResourceOrThrow

mposa:
- Adds a Python binding for `common/FindResourceOrThrow`
- Adds Python bindings to Director dependences
- Creates a bundled Bazel dependency for all `pydairlib` bindings


bindings/pydairlib/common_py.cc
```diff
@@ -0,0 +1,25 @@
...
+#include <pybind11/stl.h>
+
+#include "common/find_resource.h"
+
+namespace py = pybind11;
+
+namespace dairlib  {
+namespace pydairlib {
+
+PYBIND11_MODULE(common, m) {
```
yangwill, line 13: Is this file supposed to be committed? It looks like a duplicate of the one in `common/` and is not reference in the main BUILD.bazel

Issue #103, opened 2019-10-02
Data Visualization


mposa:
Per a group meeting discussion, we laid out a number of goals for data visualization and the outline of a path forward.

## Goals

- Save and load trajectory sets (a map from a string name to a saved trajectory object).
  - Each saved trajectory would be
    - A VectorXd of times
    - A MatrixXd of data points
    - A vector of strings, naming the rows of the data matrix
    - The string name of the trajectories (also the map key)
  - A saved trajectory set would also include meta data
    - Date/time
    - Name
    - Description
    - Git commit number (to be parsed automatically)
    - Git dirty flag (parsed automatically, true if there are uncommitted changes)
- Visualize a saved trajectory
  - Maybe eventually visualize multiple trajectories
- Visualize single and multiple saved states
- Visualize forward kinematic points (as points) and vectors (as arrows)
- Trace forward kinematic paths over trajectories
- Plot trajectories (matplotlib)
- Plot forward kinematic trajectories

@yangwill is looking into trajectory save/load
@mshalm is looking into Drake's SceneGraph

yminchen:
Just throwing out a idea: visualizing the planned trajectory online (e.g. COM trajectory, swing foot trajectory) like what Agility Robotics does in this video: https://youtu.be/2CYOrkwJzag

mposa:
Some of this list has been addressed. I've updated the main comment accordingly. @yminchen @Jianshu-Hu @yangwill  I'm guessing you're using some mechanism to visualize saved trajectories. Is this something in master that we can point to?

yangwill:
The feature exists in master and can be seen in `run_dircon_jumping.cc` and `visualize_trajectory.cc` in `examples/Cassie/`. However it requires a decent amount of boilerplate code, so I'm working on creating a specific API for saving/loading/visualizing dircon trajectories.

PR #144, opened 2020-04-23
Add utility class to simultaneously visualize multiple robot poses and DIRCON callback.

mposa:
Related to #103 

Visualizes multiple states simultaneously, also wraps into a DIRCON callback function.

Parses a URDF multiple times to achieve this, with the only option of welding one frame to the world. It would probably be better to simple work off of an existing MBP, if that were possible.

### DIFF
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -633,6 +633,8 @@ void DoMain(double duration, int max_iter, string data_directory,
     }
   }
 
+  trajopt->CreateVisualizationCallback("examples/Cassie/urdf/cassie_v2.urdf");
+
   cout << "\nChoose the best solver: "
        << drake::solvers::ChooseBestSolver(*trajopt).name() << endl;
 
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -634,7 +634,7 @@ void DoMain(double duration, int max_iter, string data_directory,
   }
 
   trajopt->CreateVisualizationCallback(
-      "examples/Cassie/urdf/cassie_fixed_springs.urdf");
+      "examples/Cassie/urdf/cassie_fixed_springs.urdf", 5);
 
   cout << "\nChoose the best solver: "
        << drake::solvers::ChooseBestSolver(*trajopt).name() << endl;
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -633,7 +633,8 @@ void DoMain(double duration, int max_iter, string data_directory,
     }
   }
 
-  trajopt->CreateVisualizationCallback("examples/Cassie/urdf/cassie_v2.urdf");
+  trajopt->CreateVisualizationCallback(
+      "examples/Cassie/urdf/cassie_fixed_springs.urdf");
 
   cout << "\nChoose the best solver: "
        << drake::solvers::ChooseBestSolver(*trajopt).name() << endl;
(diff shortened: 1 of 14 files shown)


RussTedrake:
Nice.  Related to https://github.com/RobotLocomotion/drake/issues/9728 .

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -548,6 +548,28 @@ void HybridDircon<T>::ScaleKinConstraintSlackVariables(
...
+  auto my_callback = [this](const Eigen::Ref<const VectorXd>& vars) {
+    VectorXd vars_copy = vars;
+    Eigen::Map<MatrixXd> states(vars_copy.data(),
+        this->plant_.num_positions() + plant_.num_velocities(), this->N());
+
+    this->callback_visualizer_->DrawPoses(
+        states.block(0, 0, plant_.num_positions(), this->N()));
+  };
+
+  auto binding = AddVisualizationCallback(my_callback, x_vars());
```
yminchen, line 570: This binding doesn't seem to be used, so we can remove the assignment?

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -633,6 +633,8 @@ void DoMain(double duration, int max_iter, string data_directory,
     }
   }
 
+  trajopt->CreateVisualizationCallback("examples/Cassie/urdf/cassie_v2.urdf");
```
yminchen, line 636: `cassie_v2.urdf` should be replaced by `cassie_fixed_springs.urdf`

yminchen: Additionally, we can add `CreateVisualizationCallback` to the `run_dircon_walking` example (line 908) as well

systems/trajectory_optimization/hybrid_dircon.cc
mposa: Done.

examples/Cassie/run_dircon_squatting.cc
mposa: Done.

mposa:
I could also add an option to not show every knot point during TrajectoryOptimization (instead showing every T knot points). Would this be useful?

mposa:
Will also add transparency as an option, once upstream Drake feature is implemented. 

https://github.com/RobotLocomotion/drake/issues/13126

yminchen:
> I could also add an option to not show every knot point during TrajectoryOptimization (instead showing every T knot points). Would this be useful?

Yes this feature will be helpful, especially when we cannot change the transparency of the poses between start and end. 
I'm imagining after https://github.com/RobotLocomotion/drake/issues/13126 is fixed we can show non-transparent poses every T knot points, and show the rest of the poses with high transparency.

yminchen:
> I'm imagining after [RobotLocomotion/drake#13126](https://github.com/RobotLocomotion/drake/issues/13126) is fixed we can show non-transparent poses every T knot points, and show the rest of the poses with high transparency.

Actually, maybe showing the rest of the poses with high transparency is not a good idea, since the poses are probably going to overlap a lot and we won't gain any information from most of the poses.

mposa:
Added the option to display fractional poses, which I think is worthwhile. Note that it shows poses

(1, 1 + T, 1 + 2T, 1 + 3T, ..., N)

always showing the last pose. Take a look and see what you think.

yminchen:
Another use case just occurred to me, I think it will be great to have an API which takes a list of the key frame indices? For example, if we have a multi-mode trajectory, and the key frames we want to see might not necessarily locate at `(1, 1 + T, 1 + 2T, 1 + 3T, ..., N)`

yangwill:
LGTM! Having the option to specify the number of poses when calling `DrawPoses()` might be slightly more flexible but I don't see a dire need for that in the foreseeable future.

By the way, I have code in a branch to visualize the trajectory after saving the states at the knot points to a LCM trajectory, so you don't need to rerun the optimization problem just to visualize something. Could be useful later on.

PR #148, opened 2020-05-27
Nonlinear constraint

mposa:
Move DirconAbstractConstraint to its own non-dircon file with a new name. Minor code cleanup.


mposa:
Bad files were added here. Will re-create.

PR #149, opened 2020-05-27
Moved DirconAbstractConstraint out of DIRCON and into its own class, …

mposa:
Move DirconAbstractConstraint to its own non-dircon file with a new name. Minor code cleanup.


yminchen:
Looks good to me

mposa:
Thanks for that fix--it built on my local Drake copy which is mid-update to Drake master. That include must have been pulled in by something else. Will merge once CI finishes.

PR #153, opened 2020-05-28
Minor update to MBP OSC

yminchen:
I didn't notice these changes were not updated to the MBP version of TrackingData when reviewing the PR yesterday 

Changes: 
* More detailed documentation
* Change dy to ydot (and ddy to yddot)


yangwill:
Oh I must have forgotten to add the changes to the OSC from my other branch. I have another change to the lcmt_osc_output that I also want to add, but I'll just do that in another PR

PR #155, opened 2020-05-28
Kinematic evaluator

mposa:
Per discussion earlier today, adds KinematicEvaluator abstract class and subclasses for planar contact evaluation and distance evaluations.


mposa:
Requesting review from either @yminchen  or @yangwill

mposa:
Probably good enough to review now. The next PR will use these to find fixed points.

yangwill:
Yeah i think that that would be useful, as long as it's still easy to access the values for phi, J, etc for a particular kinematic calculation for debugging purposes.

yminchen:
Yes, I think it's nice to have KinematicEvaluatorSet. 

One thing that Dircon's `KinematicsDataSet` is lacking is the ability to identify the repeated `KinematicsData`s between two given `KinematicsDataSet`s. Now that we are building a new class from scratch, do you think it's possible to add this feature in? 
It could look something like: 
```
set_idx_list = identify_repeated_set(set_1, set_2)
```
After having this `set_idx_list`, we can disable to the corresponding `phi`, 'J', and 'JdotV'

mposa:
Pushed such a set, let me know what you think of the API.

multibody/kinematic/kinematic_evaluator.cc
```diff
@@ -0,0 +1,120 @@
...
+void KinematicEvaluator<T>::set_active_inds(std::vector<int> active_inds) {
+  // TODO: should probably check for uniqueness
+  active_inds_ = active_inds;
+  num_active_ = active_inds_.size();
+
+  // Cannot just check length, because active_inds could be out of order
+  all_active_ = true;
+  if (num_active_ == length_) {
+    for (int i = 0; i < num_active_; i++) {
+      if (active_inds_.at(i) != i) {
```
yangwill, line 107: It might be helpful to also have a check that none of the active_inds are greater than length_

multibody/kinematic/kinematic_evaluator_set.h
```diff
@@ -0,0 +1,85 @@
...
+  int evaluator_active_start(int index) const;
+
+  KinematicEvaluator<T>* get_evaluator(int index) {
+    return evaluators_.at(index);
+  };
+
+  /// Adds an evaluator to the end of the list, returning the associated index
+  int add_evaluator(KinematicEvaluator<T>* e);
+
+  /// Count the total number of active constraints
```
yangwill, line 72: Probably shouldn't be called constraints here, maybe "active rows"?

multibody/kinematic/test/kinematic_evaluator_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  Jdotv_expected << 0, 0, 6.5;
+  EXPECT_TRUE(CompareMatrices(Jdotv, Jdotv_expected, tolerance));
+
+  auto Jdotv_active = evaluator.EvalActiveJacobianDotTimesV(*context);
+  VectorXd Jdotv_active_expected(1);
+  Jdotv_active_expected << 6.5;
+  EXPECT_TRUE(CompareMatrices(Jdotv_active, Jdotv_active_expected, tolerance));
+
+  // Non-default evaluator: non-vertical normal, non-zero offset
+  // Performing minmal tests
```
yangwill, line 110: tiny typo in "minimal"

multibody/kinematic/test/kinematic_evaluator_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  Vector3d pt_B({0, 0, -.5});
+  const auto& frame_B = plant_->GetFrameByName("left_lower_leg");
+
+  // Default evaluator: vertical normal, zero offset
+  auto evaluator = DistanceEvaluator<double>(*plant_, pt_A, frame_A, pt_B,
+      frame_B, .5);
+
+  auto context = plant_->CreateDefaultContext();
+
+  // Test q = 0, legs straight down
```
yangwill, line 135: typo: should be q = pi/2.

multibody/kinematic/kinematic_evaluator.cc
mposa: Done. Also checking < 0.

multibody/kinematic/kinematic_evaluator.h
```diff
@@ -0,0 +1,93 @@
...
+#include "drake/multibody/plant/multibody_plant.h"
+#include "drake/systems/framework/context.h"
+
+namespace dairlib {
+namespace multibody {
+
+/// Virtual class to represent arbitrary kinematic evaluations
+/// Evaluations are defined by some function phi(q). Implementations
+/// must generate phi(q), J(q) (the Jacboian w.r.t. velocities v), 
+/// d/dt phi(q), and [d/dt J] * v
```
yminchen, line 12: Probably a typo here. `d/dt phi(q)` seems to be implemented in `EvalFullTimeDerivative`, so the derived classes don't need to implement it

multibody/kinematic/kinematic_evaluator.cc
```diff
@@ -0,0 +1,125 @@
...
+}
+
+template <typename T>
+void KinematicEvaluator<T>::set_active_inds(std::vector<int> active_inds) {
+  // Check active_inds [0, length()] bounds
+  for (auto& i : active_inds) {
+    DRAKE_DEMAND(i >= 0);
+    DRAKE_DEMAND(i < num_full());
+  }
+  // TODO: should probably check for uniqueness
```
yminchen, line 104: I agree that we should check uniqueness. By a quick google search, this piece of code remove the duplicates: 
```
std::sort( vec.begin(), vec.end() );
vec.erase( std::unique( vec.begin(), vec.end() ), vec.end() );
```

multibody/kinematic/kinematic_evaluator.cc
```diff
@@ -0,0 +1,125 @@
...
+  active_inds_ = active_inds;
+  num_active_ = active_inds_.size();
+
+  // Cannot just check length, because active_inds could be out of order
+  all_active_ = true;
+  if (num_active_ == length_) {
+    for (int i = 0; i < num_active_; i++) {
```
yminchen, line 111: It looks like the code here is trying to detect whether the list contains all indices? If so, I think the algorithm here only works if we sort the list first.

multibody/kinematic/kinematic_evaluator_set.h
```diff
@@ -0,0 +1,87 @@
...
+  /// Determines the list of evaluators contained in the union with another set
+  /// Specifically, `index` is in the returned vector if
+  /// other.evaluators_.at(index) is an element of other.evaluators, as judged
+  /// by a comparison of the KinematicEvaluator<T>* pointers.
+  ///
+  /// Again, note that this is an index set into the other object, not self.
+  std::vector<int> FindUnion(KinematicEvaluatorSet<T> other);
+
+  /// Gets the starting index into phi_full of the specified evaluator
+  int evaluator_full_start(int index) const;
```
yminchen, line 61: Does it make more sense to use `KinematicEvaluatorSet<T>` as the input here? (same question for the function next to this)

multibody/kinematic/distance_evaluator.h
```diff
@@ -0,0 +1,52 @@
...
+#include "drake/systems/framework/context.h"
+
+namespace dairlib {
+namespace multibody {
+
+/// Basic contact evaluator for the distance between point A in frame A
+/// and point B in frame B
+///   phi(q) = ||pt_A - pt_B|| - distance
+/// Calculations are performed in B, for efficiency.
+//  The one exception is Jdotv, since Drake does not currently support
```
yminchen, line 14: nit-pick: should be ///

multibody/kinematic/world_point_evaluator.h
```diff
@@ -0,0 +1,75 @@
...
+  /// @param active_directions The directions, a subset of {0,1,2} to be
+  ///    considered active. These will correspond to rows of the rotation matrix
+  ////   Default is {0,1,2}.
+
+  WorldPointEvaluator(
+      const drake::multibody::MultibodyPlant<T>& plant,
+      const Eigen::Vector3d pt_A, const drake::multibody::Frame<T>& frame_A,
+      const Eigen::Matrix3d rotation = Eigen::Matrix3d::Identity(),
+      const Eigen::Vector3d offset = Eigen::Vector3d::Zero(),
+      std::vector<int> active_directions_ = {0,1,2});
```
yminchen, line 31: Probably typo here and the cc file: `active_directions_` -> `active_directions`

multibody/kinematic/world_point_evaluator.h
```diff
@@ -0,0 +1,75 @@
...
+  ///    the "Full" terms, but not the "Active" values
+  ///    Note that this defaults to FALSE, while the base constructor above
+  ///    includes all dimensions by default.
+
+  WorldPointEvaluator(
+      const drake::multibody::MultibodyPlant<T>& plant,
+      const Eigen::Vector3d pt_A, const drake::multibody::Frame<T>& frame_A,
+      const Eigen::Vector3d normal,
+      const Eigen::Vector3d offset = Eigen::Vector3d::Zero(),
+      bool tangent_active = false);
```
yminchen, line 54: I think it's probably better to use `active_directions` instead of `tangent_active` here so it's more consistent? 
The different between the two constructors in this class should be that one takes `Matrix3d rotation` and the other takes `Vector3d normal`. (I don't see why the second constructor disable the xy direction by default)

multibody/kinematic/test/kinematic_evaluator_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  const auto& frame_B = plant_->GetFrameByName("left_lower_leg");
+
+  // Default evaluator: vertical normal, zero offset
+  auto evaluator = DistanceEvaluator<double>(*plant_, pt_A, frame_A, pt_B,
+      frame_B, .5);
+
+  auto context = plant_->CreateDefaultContext();
+
+  // Test q = pi/2, legs straight down
+  VectorXd q(plant_->num_positions());
```
yminchen, line 136: This probably doesn't affect the test result, but I think we want to initial q with all zeros

multibody/kinematic/test/kinematic_evaluator_test.cc
```diff
@@ -0,0 +1,170 @@
...
+  const auto& frame = plant_->GetFrameByName("right_lower_leg");
+
+  // Default evaluator: vertical normal, zero offset
+  auto evaluator = WorldPointEvaluator<double>(*plant_, pt_A, frame,
+      Vector3d({0, 0, 1}), Vector3d::Zero(), false);
+
+  auto context = plant_->CreateDefaultContext();
+
+  // Test q = 0, legs straight down
+  VectorXd q(plant_->num_positions());
```
yminchen, line 62: I remember from the past experience that it is necessary to initialize the VectorXd (for examples, to zeros) manually to avoid bad initial values

multibody/kinematic/kinematic_evaluator.cc
mposa: I actually had something in here that would do this, then removed it. I'll re-add. I prefer throwing an error on non-unique solutions than secretly changing the user's request.

multibody/kinematic/kinematic_evaluator.cc
mposa: Will rename `all_active_`. It's really just a flag to determine whether the slicing operation can be skipped, only true if `active_inds_ = {0,1,2,...n}`

multibody/kinematic/kinematic_evaluator_set.h
mposa: Do you mean `KinematicEvaluator<T>*`, as the class here is `KinematicEvaluatorSet`. The code to implement that is definitely messier, so I think it's easier to implement via an integer index. Plus, since `add_evaluator` returns the index, it seems pretty easy to keep track of the indices.

If you think it would be valuable, I could add something like `GetIndex(KinematicEvaluator<T>* evaluator` that finds the appropriate index.

multibody/kinematic/world_point_evaluator.h
mposa: Ok.

multibody/kinematic/kinematic_evaluator_set.h
mposa: Scratch that. Since we're not enforcing uniqueness of the evaluators, I don't really want to add a `GetIndex` method. Will stick with integer indices.

yminchen: Ah yes, I meant `int evaluator_full_start(KinematicEvaluator<T>* e) const;`, so that the users don't need to find the index by themselves

multibody/kinematic/kinematic_evaluator.cc
yminchen: Yes throwing an error might be better

PR #156, opened 2020-05-30
Add MultibodyProgram

mposa:
Replaces the RBT fixed point and constraint solvers with a more generic MultibodyProgram. Builds off #155 and any review should wait until that is reviewed and merged. Adds a few features:

- KinematicPositionConstraint and FixedPointConstraint implementations of NonlinearConstraint, broadly usable outside of MultibodyProgram. These provide a template for how KinematicEvaluators can be used to clean up DIRCON
- MultibodyProgram as a flexible interface to MathematicalProgram for using either of the above constraints. Replaces RBT version which used multiple copy-pasta implementations of "Solver" classes
- Constraint factory classes to automatically generate friction cone constraints, usable here and elsewhere.
- cassie_utils methods to generate kinematic details (DistanceEvaluators for loop closure and more flexible std::pairs for the contact points on the feet). These should be useful in eliminating some serious copy-pasta throughout Cassie examples
- An example of finding fixed points for the fixed-springs version of Cassie. For all feasible heights, finds a solution in ~50-100ms on my laptop.

Not in this PR:
Full replacement for ContactToolkit, specifically CalcTimeDerivatives(context, lambda) and CalcMVDot(context,lambda). This will be next.


mposa:
This is ready for review.

examples/Cassie/cassie_fixed_point_solver.cc
```diff
@@ -0,0 +1,165 @@
...
+#include "drake/solvers/snopt_solver.h"
+
+namespace dairlib {
+using Eigen::VectorXd;
+void CassieFixedPointSolver(std::string filename, double height, double mu,
+    double min_normal_force, bool linear_friction_cone, VectorXd* q_result,
+    VectorXd* u_result, VectorXd* lambda_result, bool draw_pose) {
+  // Build plant
+  drake::multibody::MultibodyPlant<double> plant(0);
+  drake::multibody::MultibodyPlant<double> plant2(0);
```
yangwill, line 21: Why are there two plants? plant2 looks to be unused

examples/Cassie/cassie_fixed_point_solver.cc
```diff
@@ -0,0 +1,165 @@
...
+      Eigen::Vector3d::Zero(), false);
+  evaluators.add_evaluator(&right_heel_evaluator);
+
+  auto program = multibody::MultibodyProgram(plant, evaluators);
+
+  auto positions_map = multibody::makeNameToPositionsMap(plant);
+  auto q = program.AddPositionVariables();
+  auto u = program.AddInputVariables();
+  auto lambda = program.AddConstraintForceVariables();
+  auto kinematic_constraint = program.AddKinematicConstraint(q);
```
yangwill, line 68: The workflow of constructing the MathematicalProgram with the kinematic evaluators, and then later calling AddKinematicConstraint() with only the position decision variables as the parameter is not easy to read. Would it be possible to adjust the API to have MultibodyProgram be constructed using only the MBP and have AddKinematicConstraint() take in both q and evaluators?

examples/Cassie/cassie_utils.cc
```diff
@@ -97,6 +99,60 @@ std::unique_ptr<RigidBodyTree<double>> makeCassieTreePointer(
...
+
+template <typename T>
+std::pair<const Vector3d, const Frame<T>&> RightHeel(
+    const MultibodyPlant<T>& plant) {
+  return std::pair<const Vector3d, const Frame<T>&>(
+    Vector3d(0.088, 0, 0), plant.GetFrameByName("toe_right"));
+}
+
+template <typename T>
+multibody::DistanceEvaluator<T> LeftLoopClosureEvaluator(
```
yangwill, line 131: I'm glad there's finally a function for these constraints!

examples/Cassie/find_fixed_point.cc
```diff
@@ -0,0 +1,35 @@
...
+int do_main(int argc, char* argv[]) {
+  gflags::ParseCommandLineFlags(&argc, &argv, true);
+  srand((unsigned int) time(0));
+  drake::logging::set_log_level("err");  // ignore warnings about joint limits
+
+  std::string urdf = "examples/Cassie/urdf/cassie_fixed_springs.urdf";
+
+  Eigen::VectorXd q, u, lambda;
+
+  CassieFixedPointSolver(urdf, FLAGS_height, FLAGS_mu, 50,
```
yangwill, line 21: nitpick, make 50 a variable for readability

multibody/multibody_solvers.cc
```diff
@@ -0,0 +1,174 @@
...
+    const auto& joint = plant_.get_joint(drake::multibody::JointIndex(i));
+    if (joint.num_positions() > 0) {
+      auto q_joint = q.segment(joint.position_start(), joint.num_positions());
+      AddConstraint(q_joint <= joint.position_upper_limits());
+      AddConstraint(q_joint >= joint.position_lower_limits()); 
+  }
+  }
+}
+
+
```
yangwill, line 77: extra spaces here

multibody/multibody_solvers.cc
```diff
@@ -0,0 +1,174 @@
...
+Binding<Constraint> MultibodyProgram<T>::AddKinematicConstraint(
+    VectorXDecisionVariable q) {
+  auto constraint = std::make_shared<KinematicPositionConstraint<T>>(
+        plant_, evaluators_, context_);
+  return AddConstraint(constraint, q);
+}
+
+template <typename T>
+Binding<Constraint> MultibodyProgram<T>::AddFixedPointConstraint(
+    VectorXDecisionVariable q, VectorXDecisionVariable u,
```
yangwill, line 57: q, u, and lambda can be const refs

multibody/contact_toolkit.h
```diff
@@ -12,17 +12,24 @@ namespace multibody {
 /*
  * ContactInfo structure that holds all the required contact information.
  * xA: Positions on Body A (The robot) expressed in the frame of the robot.
- * xB: Positions on Body B (Assumed to be the ground) expressed in the ground
  * frame (world frame).
```
yangwill, line 15: frame is no longer a parameter

multibody/contact_toolkit.h
```diff
@@ -12,17 +12,24 @@ namespace multibody {
...
- * idxA: Body indices of the corresponding positions in the RigidBodyTree.
+ * frameA: Frame of the corresponding positions in the MultibodyPlant.
```
yangwill, line 16: xA and frameA are not very informative variable names. How about we rename xA to contact_point, and frameA to contact_frame?

examples/Cassie/cassie_fixed_point_solver.cc
mposa: I'm not sure I did this the right way, so I'm happy to have this discussion. Right now, the code straddles a weird place where, because it extends MathematicalProgram, it's trying to "own" the plant/evaluators/decision variables/etc. However, I don't think it's doing so particularly effectively for the reason you stated. If I remove owning the evaluators, given that ownership of the decision variables is already a fraud, then the only thing in the constructor would be plant itself. At this point, is MultibodyProgram really a class?

These could, instead, just be global utility methods.
```
DecisionVariables  q = AddPositionVariables(program*, plant&)
...
DecisionVariables lambda = AddConstraintForceVariables(program*, plant&)
Binding = AddKinematicConstraint(program*, plant&, evaluators&, q)
Binding = AddFixedPointConstraint(program*, plant&, evaluators&, q, u, lambda)
```

yangwill: I prefer keeping MultibodyProgram as a class, it has enough utility (the various AddConstraint functions that only make sense when they refer to a MBP) that it's worth keeping them together as a single class rather than as separate utility methods.

I would prefer an API that looked like:

```
DecisionVariables q = program.AddPositionVariables()
Binding = program.AddKinematicConstraint(evaluators&, q)
```

That way it's easy to see the dependencies.

examples/Cassie/cassie_fixed_point_solver.cc
```diff
@@ -0,0 +1,164 @@
...
+#include "multibody/kinematic/world_point_evaluator.h"
+#include "multibody/multibody_solvers.h"
+#include "multibody/multibody_utils.h"
+#include "multibody/multipose_visualizer.h"
+#include "solvers/constraint_factory.h"
+
+#include "drake/common/text_logging.h"
+#include "drake/multibody/parsing/parser.h"
+#include "drake/solvers/solve.h"
+#include "drake/solvers/snopt_solver.h"
```
yminchen, line 12: nit: `snopt_solver.h` is not used in this file

multibody/multibody_solvers.h
```diff
@@ -0,0 +1,110 @@
...
+  drake::solvers::VectorXDecisionVariable q_;
+  drake::solvers::VectorXDecisionVariable u_;
+  drake::solvers::VectorXDecisionVariable lambda_;
+};
+
+template <typename T>
+class KinematicPositionConstraint : public solvers::NonlinearConstraint<T> {
+ public:
+  /// This constructor takes a shared_ptr<Context> as an argument to share
+  /// cached kinematic/dynamic computation within the ccontext
```
yminchen, line 61: typo: context

multibody/multibody_solvers.cc
```diff
@@ -0,0 +1,171 @@
...
+  return AddConstraint(constraint, {q, u, lambda});
+}
+
+template <typename T>
+void MultibodyProgram<T>::AddJointLimitConstraints(VectorXDecisionVariable q) {
+  for (int i = 0; i < plant_.num_joints(); i++) {
+    const auto& joint = plant_.get_joint(drake::multibody::JointIndex(i));
+    if (joint.num_positions() > 0) {
+      auto q_joint = q.segment(joint.position_start(), joint.num_positions());
+      AddConstraint(q_joint <= joint.position_upper_limits());
```
yminchen, line 71: Just to confirm, do the joint limits work in the case of continuous time MBP? I remember seeing warnings about joint limits are not supportet in this case.

multibody/kinematic/test/kinematic_evaluator_caching_test.cc
```diff
@@ -0,0 +1,257 @@
+#include <chrono>
+
+#include "examples/Cassie/cassie_utils.h"
+#include "multibody/kinematic/kinematic_evaluator_set.h"
+
+#include "drake/common/text_logging.h"
+#include "drake/multibody/parsing/parser.h"
+
+/// This is a timing test to determine how effective caching is using
+/// MultibodyPlant kinematics cacheing via a context.
```
yminchen, line 10: Typo: caching

multibody/kinematic/test/kinematic_evaluator_caching_test.cc
```diff
@@ -0,0 +1,257 @@
...
+/// Tests phi, J, Jdotv in a few scenarios
+/// Results show that kinematics effectively track dependencies, and setting
+///  context.q = context.q
+/// will not make the cache dirty.
+///
+/// As of 5/31/2020, phi, J make effective use of the cache, across bodies,
+/// but Jdotv is ony partially effective.
```
yminchen, line 17: Typo: only

examples/Cassie/cassie_fixed_point_solver.cc
yminchen: To add my point of view. I think it's pretty easy to read `MultibodyProgram`, but I admit I expected that `AddKinematicConstraint` takes in `evaluators` as an argument, the first time I saw the function. 

If we are going to have `AddKinematicConstraint(evaluators, q)`, then we will need to make sure that the users pass the same `evaluators` into both  `AddKinematicConstraint` and `AddFixedPointConstraint`. I don't have a clean solution to implement this right now

mposa: I agree with both of you, and have a new idea that might satisfy everyone. The basic question, I think, is whether or not a `MultibodyProgram` is allowed have two evaluator sets, or multiple robot states/inputs/forces. If the answer is "no," then we can take the current implementation and _remove_ the `q,u,lambda` inputs to the `AddXXConstraint()` functions, and just DEMAND that they've been created.

However, if the answer is "yes" then the API becomes cleaner. Like Will suggested, no `evaluators` as a constructor argument. The program will no longer keep references to `q,u,lambda`, as these may be non-unique. Instead, `AddFixedPointConstraint(evaluators,q,u,lambda)` will confirm that the lengths of `q,u,lambda` are appropriate for the plant and evaluators. Similar for `AddKinematicConstraint`.

Everything becomes a bit more flexible, and the restrictions on the arguments are that 

1. evaluators.plant() matches self.plant()
2. The lengths of the input decision variables are appropriate.

Thoughts?

multibody/multibody_solvers.cc
mposa: You're right, they are not used when _simulating_ the continuous time MBP. However, they still exist and can be referenced.

examples/Cassie/cassie_fixed_point_solver.cc
yminchen: From what I can imagine, we probably don't need multiple states/inputs/forces and evaluator sets. People could just create multiple `MultibodyProgram`s to achieve this?

And I have a question about the proposed structure, do we demand that the evaluators passed into `AddKinematicConstraint` and `AddFixedPointConstraint` are the same?

multibody/multibody_solvers.cc
yminchen: Got it!

examples/Cassie/cassie_fixed_point_solver.cc
mposa: This framework would flexible enough to optimize for two different states, where those states are related somehow.

Via the proposed structure, that would not be enforced--since there's really no need to.

mposa:
Recent testing on the next branch has revealed some memory issues here that did not crop up in initial testing. However, to avoid a messy merge, will address in the next PR.

PR #157, opened 2020-06-02
Contact toolkit replacement

mposa:
This is a relatively heavy PR with a few items of note.

- Deletes ContactToolkit, replaces with a variety of methods in KinematicConstraintSet
- Tweaks CassieFixedPointSolver. No longer as reliably fast, but allows option to spread toes out.
- Adds options to cassie_utils MBP construction to include/exclude springs and loop closure springs
- Uses CassieFixedPointSolver for run_dircon_squatting. Slows down the result (52->72 seconds), likely due to a less finely-tuned initial guess. Did not try to improve.
- Fix what seems to be a memory leak in MultibodyProgram related to the context shared pointer. Not totally clear what the cause was.
- Added constrained LQR controller. Works well on Cassie where springs are fixed. Works less well on the spring model, likely due to model stiffness. Did not try for long to improve via tuning. Works best for height=1 and requires a small simulation dt.

Note--a number of Cassie-examples (particularly the EKF) use attic/contact_toolkit and attic/multibody_solvers, thus these cannot be removed yet.


examples/Cassie/multibody_sim.cc
```diff
@@ -80,11 +77,20 @@ int do_main(int argc, char* argv[]) {
...
 
+  std::string urdf;
+  if (FLAGS_spring_model) {
+    urdf = "examples/Cassie/urdf/cassie_v2.urdf";
+  } else {
+    urdf = "examples/Cassie/urdf/cassie_fixed_springs.urdf";
+  }
+
+  addCassieMultibody(&plant, &scene_graph, FLAGS_floating_base, urdf,
+      FLAGS_spring_model, true);
```
yminchen, line 89: I'm not sure if this is done intentionally. The last should be `FLAGS_spring_model`?

yminchen: Ah never mind, I was misled by the argument name. I thought the it is a flag for the heel leaf springs.

yminchen: Maybe we can change the name to `add_loop_closure` instead of `add_loop_closure_springs`?

examples/Cassie/cassie_utils.h
```diff
@@ -41,10 +41,15 @@ using drake::systems::lcm::LcmPublisherSystem;
...
+/// @param add_leaf_springs Whether or not to add the 4 leaf springs in the legs
+///     Default = true
+/// @param add_loop_closure_springs Whether or not to add the loop closure
+///     distance constraint via stiff springs. Default = true.
 void addCassieMultibody(drake::multibody::MultibodyPlant<double>* plant,
     drake::geometry::SceneGraph<double>* scene_graph = nullptr,
     bool floating_base = true,
-    std::string filename = "examples/Cassie/urdf/cassie_v2.urdf");
+    std::string filename = "examples/Cassie/urdf/cassie_v2.urdf",
+    bool add_leaf_springs = true,  bool add_loop_closure_springs = true);
```
yminchen, line 52: Should we throw an error if 
filename = "examples/Cassie/urdf/cassie_fixed_spring.urdf
and 
add_leaf_springs = true?

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -405,8 +227,8 @@ void DoMain(double duration, int max_iter, string data_directory,
       plant, num_time_samples, min_dt, max_dt, dataset_list, options_list);
 
   // Snopt settings
-  //   trajopt->SetSolverOption(drake::solvers::SnoptSolver::id(),
-  //                            "Print file", "../snopt.out");
+  trajopt->SetSolverOption(drake::solvers::SnoptSolver::id(),
```
yminchen, line 230: Default to no output file?

examples/Cassie/urdf/cassie_v2.urdf
```diff
@@ -714,7 +714,7 @@
     <child link="shin_left" />
     <axis xyz="0 0 1" />
     <limit lower="-.1" upper="0.075"/>
-    <dynamics damping=".10" />
+    <dynamics damping="10" />
```
yminchen, line 717: I'm curious why the damping is 10?

multibody/kinematic/kinematic_evaluator_set.cc
```diff
@@ -134,6 +135,106 @@ std::vector<int> KinematicEvaluatorSet<T>::FindUnion(
...
+  VectorX<T> tau_g = plant_.CalcGravityGeneralizedForces(context);
+
+  drake::multibody::MultibodyForces<T> f_app(plant_);
+  plant_.CalcForceElementsContribution(context, &f_app);
+
+  // Evaluate constraint terms, phi/phidot for active. Jacboians for full
+  VectorX<T> phi = EvalActive(context);
+  VectorX<T> phidot = EvalActiveTimeDerivative(context);
+  MatrixX<T> J = EvalFullJacobian(context);
+  VectorX<T> Jdotv = EvalFullJacobianDotTimesV(context);
```
yminchen, line 214: Shouldn't the dimensions of `Jdotv`, `phi`, `phidot` be the same according to
  // [[M -J^T]  [[vdot  ]  =  [[tau_g + Bu - C     ]
  //  [J  0 ]]  [lambda]]     [ -Jdotv - kp phi - kd phidot]]
?

multibody/kinematic/kinematic_evaluator_set.cc
```diff
@@ -134,6 +135,106 @@ std::vector<int> KinematicEvaluatorSet<T>::FindUnion(
...
+VectorX<T> KinematicEvaluatorSet<T>::CalcTimeDerivatives(
+    const Context<T>& context, double alpha) const {
+  VectorX<T> lambda;
+  return CalcTimeDerivatives(context, &lambda, alpha);
+}
+
+template <typename T>
+VectorX<T> KinematicEvaluatorSet<T>::CalcTimeDerivatives(
+    const Context<T>& context, VectorX<T>* lambda, double alpha) const {
+  // M(q) vdot + C(q,v) = tau_g(q) + Bu + J(q)^T lambda
```
yminchen, line 189: Missing the term `F_app` here and three lines below

multibody/multibody_solvers.h
```diff
@@ -52,7 +52,7 @@ class MultibodyProgram : public drake::solvers::MathematicalProgram {
   drake::solvers::Binding<drake::solvers::Constraint> AddKinematicConstraint(
       const KinematicEvaluatorSet<T>& evaluators,
       const drake::solvers::VectorXDecisionVariable& q,
-      std::shared_ptr<drake::systems::Context<T>> local_context);
+      drake::systems::Context<T>* local_context);
```
yminchen, line 55: I don't see `AddKinematicConstraint(evaluators, q, local_context)` being used anywhere in dairlib. Are we going to use the local_context version of AddXXXConstraint?

systems/controllers/constrained_lqr_controller.h
```diff
@@ -38,12 +40,11 @@ class ConstrainedLQRController : public drake::systems::LeafSystem<double> {
    * The constructor computes K and E and is the major computational block of
    * the controller class.
    */
-  ConstrainedLQRController(const RigidBodyTree<double>& tree,
-                           const Eigen::VectorXd& q, const Eigen::VectorXd& u,
-                           const Eigen::VectorXd& lambda,
-                           const Eigen::MatrixXd& Q, const Eigen::MatrixXd& R,
-                           const dairlib::multibody::ContactInfo& contact_info =
-                               dairlib::multibody::ContactInfo());
+  ConstrainedLQRController(
```
yminchen, line 43: Typo in line 33, provide -> provided

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -1,103 +1,95 @@
...
 
-  // Contact Jacobian (If contact information is provided)
-  if (contact_info_.num_contacts > 0) {
-    J_contact_qdot = autoDiffToValueMatrix(
-        contact_toolkit_->CalcContactJacobian(initializeAutoDiff(x), true));
-    J_contact_v = autoDiffToValueMatrix(
-        contact_toolkit_->CalcContactJacobian(initializeAutoDiff(x), false));
+  // Add quaternion constraints to F
+  // Constraints come from the kinematic fact that ||q|| = 1, linearized is
+  // dot(q_0, delta_q) = 0
```
yminchen, line 60: I didn't get what the dot(,) function means by reading it. Maybe we can change it to 
`<q_0, delta_q> = 0`
or
`q_0^T * delta_q = 0`
?

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -1,103 +1,95 @@
...
-  J_qdot << J_tree_qdot, J_contact_qdot;
-  J_v << J_tree_v, J_contact_v;
+  MatrixXd F_quat =
+      MatrixXd::Zero(num_quat, J_active_qdot.cols() + J_active_v.cols());
+  for (int i = 0; i < num_quat; i++) {
+    F_quat.row(i).segment(quat_start.at(i), 4) = autoDiffToValueMatrix(
+        plant_.GetPositions(context).segment(quat_start.at(i),4));
+  }
 
   // Computing F
```
yminchen, line 80: Maybe it's helpful to have something like this:
```
// F = [J_active_qdot, 0         ]
//     [0            , J_active_v]
//     [          F_quat         ]
```

Also, from equation (23) of the paper,  F seems to be
```
// F = [J_active_qdot, 0              ]
//     [d/dt J_active_qdot, J_active_v]
//     [            F_quat            ]
```
so the implementation here is missing a `d/dt J`?

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -107,37 +99,48 @@ ConstrainedLQRController::ConstrainedLQRController(
...
+  AutoDiffVecXd x_ad = xu_ad.head(plant_.num_positions()
+      + plant_.num_velocities());
+  AutoDiffVecXd u_ad = xu_ad.segment(plant_.num_positions()
+      + plant_.num_velocities(), plant_.num_actuators());
+
+  auto context_ad = multibody::createContext(plant_, x_ad, u_ad);
 
-  // Making sure that the derivative is zero
-  // DRAKE_DEMAND(autoDiffToValueMatrix(xdot_autodiff).isZero(1e-6));
+  AutoDiffVecXd xdot = evaluators_.CalcTimeDerivatives(*context_ad);
```
yminchen, line 118: It seems that the reason we pass the autodiff version of evaluator as a constructor's argument is because of this line and line 120. 
If we can have a method that creates `KinematicEvaluatorSet<Autodiff>` from `KinematicEvaluatorSet<double>`, then the user won't need to pass the autodiff one, and the constructor will probably be cleaner.

yminchen: I don't know if it's possible to convert Context<double> to Context<Autodiff>

systems/controllers/constrained_lqr_controller.cc
```diff
@@ -107,37 +99,48 @@ ConstrainedLQRController::ConstrainedLQRController(
...
 
-  MatrixXd AB = autoDiffToGradientMatrix(xdot_autodiff);
-  MatrixXd A = AB.leftCols(num_states_);
-  MatrixXd B = AB.block(0, num_states_, AB.rows(), num_efforts_);
+  MatrixXd AB = autoDiffToGradientMatrix(xdot);
+  MatrixXd A = AB.leftCols(plant_.num_positions() + plant_.num_velocities());
+  MatrixXd B = AB.block(0, plant_.num_positions() + plant_.num_velocities(),
+      AB.rows(), plant_.num_actuators());
```
yminchen, line 123: Probably cleaner: MatrixXd B = AB.rightCols(plant_.num_actuators());

systems/controllers/constrained_lqr_controller.cc
yminchen: I guess `d/dt J_active_qdot=0` because the nominal trajectory is stationary.

examples/Cassie/multibody_sim.cc
mposa: Done.

examples/Cassie/run_dircon_squatting.cc
mposa: Done.

examples/Cassie/urdf/cassie_v2.urdf
mposa: Will revert. These numbers, as far as I know, are not really based yet in empirical results.

multibody/kinematic/kinematic_evaluator_set.cc
mposa: Good catch! Will use active for all.

multibody/multibody_solvers.h
mposa: I think it's important to have it as an option, in case multiple constraint sets are added that correspond to multiple state decision variables (to avoid dirtying the cache when going back and forth between variable sets).

systems/controllers/constrained_lqr_controller.cc
mposa: Done

systems/controllers/constrained_lqr_controller.cc
mposa: `d/dt J = 0` is constant for the time-invariant version.

Added that helpful comment.

systems/controllers/constrained_lqr_controller.cc
mposa: Done

examples/Cassie/cassie_utils.h
mposa: At the moment, if you try that, the program will crash with the (unhelpful) error message `std::bad_cast` when this code
```
 plant->AddForceElement<RevoluteSpring>(
        dynamic_cast<const drake::multibody::RevoluteJoint<double>&>(
            plant->GetJointByName("knee_joint_left")),
        0, 1500);
```
tries to cast the fixed spring joint to a `RevoluteJoint`. However, I'm not sure how much adding the suggested check really avoids errors, given that the URDF names themselves aren't completely static.

systems/controllers/constrained_lqr_controller.cc
mposa: Agreed--but I'm not sure there's a simple way to do that (without having to use `new` and memory management). You'd essentially need to deep copy the `KinematicEvaluatorSet`, which would require copying the `KinematicEvaluator` objects--but the `Set` only stores these as pointers.

examples/Cassie/cassie_utils.h
yminchen: Got it, maybe we can add a warning to the documentation?

multibody/kinematic/kinematic_evaluator_set.h
```diff
@@ -57,6 +57,50 @@ class KinematicEvaluatorSet {
...
+      const drake::VectorX<T>& lambda) const;
+
+  /// Computes vdot given the state, control inputs and constraint
+  /// forces. Similar to CalcMassMatrixTimesVDot, but uses inv(M)
+  /// and includes qdot. Forces are associated with the full kinematic
+  /// elements.
+  /// @param context
+  /// @param lambda constraint forces, applied via
+  ///   evaluators.EvalActiveJacobian().transpose() * lambda
+  drake::VectorX<T> CalcTimeDerivativesGivenForce(
```
yangwill, line 76: GivenForce sounds weird. Depending on which function you anticipate being called more often, it might make more sense to rename the function with the user supplied lambda CalcTimeDerivatives(), and rename the current CalcTimeDerivatives() to be CalcTimeDerivativesAndForces()

Reading it a second time, it makes more sense. Make rename it to CalcTimeDerivativesWithForce() or CalcTimeDerivativesUsingForce()?

systems/controllers/constrained_lqr_controller.h
```diff
@@ -28,7 +30,7 @@ namespace systems {
  * obtain q*, u* and lambda* (Fixed point vectors).
  * The Q and R matrices need to be specified, which define the LQR cost
  * matrices.
- * If the system is in contact, ContactInfo may be provide and this information
+ * If the system is in contact, ContactInfo may be provided and this information
```
yangwill, line 33: ContactInfo doesn't seem to be used anymore

multibody/kinematic/kinematic_evaluator_set.h
mposa: Done.

Issue #37, opened 2019-02-19
MultibodyPlant IMU sensors


mposa:
Add IMU (accelerometer and gyroscope) sensors to MultibodyPlant and contribute back to Drake.

mposa:
Resolved via https://github.com/RobotLocomotion/drake/pull/13501 and https://github.com/RobotLocomotion/drake/pull/13604 and

PR #159, opened 2020-06-03
Fix procman script

yminchen:
Fix a bug in procman script

### DIFF
--- a/examples/Cassie/cassie_simulation.pmd
+++ b/examples/Cassie/cassie_simulation.pmd
@@ -80,7 +80,11 @@ group "0.operator" {
         exec = "bazel-bin/examples/Cassie/visualizer --floating_base=true";
         host = "localhost";
     }
-    cmd "1.state-visualizer-fixed-base" {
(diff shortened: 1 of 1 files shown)


yangwill:
Looks fine, but since this is updating the procman config anyway can we also change the simulator to use the new MBP simulator?

yminchen:
> Looks good to me, but since this is updating the procman config anyway can we also change the simulator to use the new MBP simulator?

I replaced `run_simple_sim` with `multibody_sim`, but there is a incompatibility in the case of fixed-based Cassie. I didn't dig into what caused the error, but the bug doesn't seem to exist in Michael's current PR. I also noticed that either case doesn't have `lcmt_cassie_out` in `multibody_sim`, so maybe we should just wait for Michael's PR being merged into master before I make the change

mposa:
Agree--we cannot fully replace the simulation yet. We still need MBP IMU sensors #37 or https://github.com/RobotLocomotion/drake/issues/13457

PR #161, opened 2020-06-11
Created visualization script and documentation for the JSON file

JoanShaho:
The kinematics_test.py contains the working visualization script. The testJSON.json is an example JSON file and the JSON_Visualization_README.txt explains how the JSON file is structured.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/161)

### DIFF
--- a/director/scripts/test/JSON_Visualization_README.txt
+++ b/director/scripts/test/JSON_Visualization_README.txt
@@ -23,7 +23,11 @@ visualization of data during a simulation:
         "alpha": The transparency as a value from 0 to 1
         "type": "line",
         "thickness": The thickness of the line
-        "history": The time (in seconds) that the trace line will cover
+        "history": The time (in seconds) that the trace line will cover. Note
+                   that this assumes that there is a constant rate of messages
+                   coming through. However, if the messages are not coming at a
+                   constant rate then the displayed line might be of varying
+                   length.
 
         2) A point which has the following properties:
 
--- a/director/scripts/test/JSON_Visualization_README.txt
+++ b/director/scripts/test/JSON_Visualization_README.txt
@@ -1,0 +1,36 @@
+This documentation explains the different parts of the JSON file needed for the
+visualization of data during a simulation:
+
+"model_file": This is the directory containing the file for the robot/plant
+              description (like an urdf file).
+
+"weld-body": This is whether the user wants "WeldFrames()" to be called on the
+             robot/plant. If this property does not exist then the "WeldFrames()"
+             function will not be called
+
+"data": This will be a list of all the different visualizations/massages that
+        the user wants to be displayed. The 3 types of visualizations will be
+        kinematic data, the center of mass, and an LCM message. So far there
+        are 2 types of visualizations that can be shown both of which are
+        kinematic data:
+
+        1) A trace line which has the following properties:
+
+        "name": The line's name to be displayed in the GUI
+        "body_part": The name of the body part/context on which the line would be
+        "point-on-body": The location within the body part/context
+        "color": The line's color in the format [Red, Green, Blue]
+        "alpha": The transparency as a value from 0 to 1
+        "type": "line",
+        "thickness": The thickness of the line
+        "history": The time (in seconds) that the trace line will cover
+
+        2) A point which has the following properties:
+
+        "name": The point's name to be displayed in the GUI
+        "body_part": The name of the body part/context on which the line would be
+        "point-on-body": The location within the body part/context
+        "color": The line's color in the format [Red, Green, Blue]
+        "alpha": The transparency as a value from 0 to 1
+        "type": "point",
+        "radius": The radius of the sphere representing the point
--- a/director/scripts/test/kinematic_test.py
+++ b/director/scripts/test/kinematic_test.py
@@ -8,6 +8,7 @@ import dairlib.lcmt_robot_output
 
 import director.applogic
 import director.mainwindowapp
+from director.debugVis import DebugData
 
 from pydairlib.common import FindResourceOrThrow
 
@@ -16,6 +17,9 @@ import pydrake.systems.framework
 import pydrake.multibody.plant
 import pydrake.multibody.parsing
 import numpy as np
+import json
+import sys
+from collections import deque
 
 
 class TestGui(QWidget):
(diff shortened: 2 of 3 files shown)


director/scripts/test/JSON_Visualization_README.txt
```diff
@@ -0,0 +1,40 @@
...
+"data": This will be a list of all the different visualizations/massages that
+        the user wants to be displayed. The 3 types of visualizations will be
+        kinematic data, the center of mass, and an LCM message. So far there
+        are 2 types of visualizations that can be shown both of which are
+        kinematic data:
+
+        1) A trace line which has the following properties:
+
+        "name": The line's name to be displayed in the GUI
+        "body_part": The name of the body part/context on which the line would be
```
mposa, line 20: suggest "frame" instead of "body_part" for this field, since that's what we actually use.

director/scripts/test/JSON_Visualization_README.txt
```diff
@@ -0,0 +1,40 @@
...
+        "point-on-body": The location within the body part/context
```
mposa, line 21: Suggest "point" since hyphens cannot be used for code variables and this is descriptive enough.

director/scripts/test/JSON_Visualization_README.txt
```diff
@@ -0,0 +1,40 @@
...
+        "history": The time (in seconds) that the trace line will cover. Note
+                   that this assumes that there is a constant rate of messages
+                   coming through. However, if the messages are not coming at a
+                   constant rate then the displayed line might be of varying
+                   length.
+
+        2) A point which has the following properties:
+
+        "name": The point's name to be displayed in the GUI
+        "body_part": The name of the body part/context on which the line would be
```
mposa, line 35: "frame"

director/scripts/test/JSON_Visualization_README.txt
```diff
@@ -0,0 +1,40 @@
...
+        "point-on-body": The location within the body part/context
```
mposa, line 36: Suggest "point"

director/scripts/test/kinematic_test.py
```diff
@@ -16,6 +17,9 @@
 import pydrake.multibody.plant
 import pydrake.multibody.parsing
 import numpy as np
+import json
+import sys
+from collections import deque
 
 
 class TestGui(QWidget):
```
mposa, line 25: Will need a real name, not TestGui

director/scripts/test/kinematic_test.py
```diff
@@ -30,21 +34,36 @@ def __init__(self, parent = None):
     def handle_thread(self):
         self.channel = "NETWORK_CASSIE_STATE_DISPATCHER"
```
mposa, line 35: State channel name should be settable(with a default) via command line, or included in the JSON file.

mposa, line 35: Everything except for the 
```
while true
            self.lcm.handle_timeout(100)
```
should probably be moved to the main thread (I know it was written like this in the test code I sent you)

director/scripts/test/kinematic_test.py
```diff
@@ -30,21 +34,36 @@ def __init__(self, parent = None):
...
         self.lcm = lcm.LCM()
+        self.prev_loc = {}
+        self.json_file = sys.argv[3];
```
mposa, line 38: This use of `sys.argv` seems like it requires this particular script to be in a specific place in the full call to this method. I suspect there's a better way to do this.

I did some (incomplete) digging, but it looks like the argument should be available in this `globalsDict` object
https://github.com/RobotLocomotion/director/blob/7139b80d3d2902a7ffdb8e3eac731c6acc68aa92/src/python/director/mainwindowapp.py#L336

I don't know how you can access it, but the name implies it's global somehow.

director/scripts/test/kinematic_test.py
```diff
@@ -56,19 +75,84 @@ def state_handler(self, channel, data):
...
-        body_name = "toe_left"
-
-        # Use Drake's CalcPointsPositions to determine where that point is
-        # in the world
-        pt_world = self.plant.CalcPointsPositions(self.context,
-            self.plant.GetFrameByName(body_name), pt_body,
-            self.plant.world_frame())
-        print(pt_world.transpose())
-        #TODO instead of printing the position, visualize it
+        for data in self.data['data']:
```
mposa, line 78: Parsing of the JSON file should occur once, before LCM handling starts. Everything needed should then be stored into convenient objects.

director/scripts/test/kinematic_test.py
```diff
@@ -30,21 +34,36 @@ def __init__(self, parent = None):
     def handle_thread(self):
         self.channel = "NETWORK_CASSIE_STATE_DISPATCHER"
         self.lcm = lcm.LCM()
+        self.prev_loc = {}
+        self.json_file = sys.argv[3];
+        self.duration = {}
```
mposa, line 39: Fields like these need to be better organized (e.g. by data type). See comment below also on JSON parsing.

director/scripts/test/kinematic_test.py
```diff
@@ -56,19 +75,84 @@ def state_handler(self, channel, data):
...
+        for data in self.data['data']:
+            jsonData = eval(str(data))
+
+            # set the body and point on which to visualize the data
+            pt_body = np.array(jsonData['point-on-body'])
+            body_name = jsonData['body_part']
+
+            # Use Drake's CalcPointsPositions to determine where that point is
+            # in the world
+            pt_world = self.plant.CalcPointsPositions(self.context,
```
mposa, line 87: These two lines could be combined

director/scripts/test/kinematic_test.py
```diff
@@ -56,19 +75,84 @@ def state_handler(self, channel, data):
...
-        pt_world = self.plant.CalcPointsPositions(self.context,
-            self.plant.GetFrameByName(body_name), pt_body,
-            self.plant.world_frame())
-        print(pt_world.transpose())
-        #TODO instead of printing the position, visualize it
+        for data in self.data['data']:
+            jsonData = eval(str(data))
+
+            # set the body and point on which to visualize the data
+            pt_body = np.array(jsonData['point-on-body'])
```
mposa, line 82: Not all types have these fields--this should be inside the switch on type.

director/scripts/test/kinematic_test.py
```diff
@@ -56,19 +75,84 @@ def state_handler(self, channel, data):
...
+            pt_world = self.plant.CalcPointsPositions(self.context,
+                self.plant.GetFrameByName(body_name), pt_body,
+                self.plant.world_frame())
+            next_loc = pt_world.transpose()[0]
+
+            # draw a continuous line
+            if (jsonData['type'] == "line"):
+
+                # check if this line has bee drawn or not
+                try:
```
mposa, line 96: Never expect failure from a try-catch. It's very computationally expensive. The way you do this for the point is better.

Issue #137, opened 2020-03-31
Redudant contact constriants in OSC


yminchen:
This issue was brought up in #95.

Currently the contact constraints are 3D per point in `OperationalSpaceControl`. This leads to potential redundant constraints. For example, the contact constraints for Cassie's foot should be 5 DoF because the foot can rotate along the line through the two contact points. 

Not sure how much this affect the performance of OSC. We will need to test it and if it helps we will merge the new code into master branch.

yminchen:
Issue resolved in PR #162

yminchen:
PR #162 only fixed the issue partially. Currently the redundant constraint direction is expressed in the world frame, but we would like to express it in local frame such as the toe frame to avoid singularity. 

Per discussion with @mposa , one quick solution could be changing the third frame to the local frame in the arguments of the following functions.
https://github.com/DAIRLab/dairlib/blob/9268cc75bac085a0829e18c9efe24400a997b5c4/multibody/kinematic/world_point_evaluator.cc#L68
https://github.com/DAIRLab/dairlib/blob/9268cc75bac085a0829e18c9efe24400a997b5c4/multibody/kinematic/world_point_evaluator.cc#L80

yminchen:
> PR #162 only fixed the issue partially. Currently the redundant constraint direction is expressed in the world frame, but we would like to express it in local frame such as the toe frame to avoid singularity.
> 
> Per discussion with @mposa , one quick solution could be using the local frame for the third frame in the arguments of the following functions.
> 
> https://github.com/DAIRLab/dairlib/blob/9268cc75bac085a0829e18c9efe24400a997b5c4/multibody/kinematic/world_point_evaluator.cc#L68
> 
> 
> https://github.com/DAIRLab/dairlib/blob/9268cc75bac085a0829e18c9efe24400a997b5c4/multibody/kinematic/world_point_evaluator.cc#L80

Simply changing the third frame to toe frame didn't work, because we express the contact force wrt the frame in OSC which means the z axis has to point to world's +z.
#267 rotates J and JdotV about global z axis (expressing J and JdotV in local frame).

PR #162, opened 2020-06-13
Port Cassie's controllers

yminchen:
Main changes: 
* Port Cassie's standing/walking controller to MBP
* Use `KinematicEvaluator` for the constraints in OSC 
* Remove redundant contact constraints in OSC (this fixes issue #137)
* Improve the implementation of multi contact modes in OSC  

Minor changes:
* Add a getter for active_inds in `KinematicEvaluator`  
* Fix a bug in OSC (original code threw segfault error if there is no contact point)

I will port EKF next


examples/Cassie/run_osc_standing_controller.cc
```diff
@@ -56,81 +57,100 @@ DEFINE_double(height, .89, "The desired height (m)");
 int DoMain(int argc, char* argv[]) {
   gflags::ParseCommandLineFlags(&argc, &argv, true);
 
+  // Build Cassie MBP
+  drake::multibody::MultibodyPlant<double> plant_w_springs(0.0);
+  addCassieMultibody(&plant_w_springs, nullptr, true /*floating base*/,
+                     "examples/Cassie/urdf/cassie_v2.urdf",
+                     true /*spring model*/, false /*loop closure*/);
```
yangwill, line 64: Why is the loop closure flag set to false?

examples/Cassie/run_osc_standing_controller.cc
```diff
@@ -56,81 +57,100 @@ DEFINE_double(height, .89, "The desired height (m)");
...
+  osc->AddContactPoint(&left_heel_evaluator);
+  auto right_toe_evaluator = multibody::WorldPointEvaluator(
+      plant_wo_springs, right_toe.first, right_toe.second, Matrix3d::Identity(),
+      Vector3d::Zero(), {1, 2});
+  osc->AddContactPoint(&right_toe_evaluator);
+  auto right_heel_evaluator = multibody::WorldPointEvaluator(
+      plant_wo_springs, right_heel.first, right_heel.second,
+      Matrix3d::Identity(), Vector3d::Zero(), {0, 1, 2});
+  osc->AddContactPoint(&right_heel_evaluator);
+  // Cost
```
yangwill, line 149: remove comments

systems/controllers/osc/operational_space_control.cc
```diff
@@ -146,28 +145,46 @@ void OperationalSpaceControl::AddAccelerationCost(
...
+  // Find the new contact in all_contacts_
+  auto it_c = std::find(all_contacts_.begin(), all_contacts_.end(), evaluator);
+  int contact_idx = std::distance(all_contacts_.begin(), it_c);
+  // Add to contact list if the new contact doesn't exist in the list
+  if (it_c == all_contacts_.cend()) {
+    all_contacts_.push_back(evaluator);
+  }
+
+  // Find the finite state machine state in contact_indices_map_
+  auto it_map = contact_indices_map_.find(state);
```
yangwill, line 167: I would recommend renaming it_map, it sounds like it's map of iterators when instead it's just an iterator

systems/controllers/osc/operational_space_control.cc
```diff
@@ -226,16 +241,24 @@ void OperationalSpaceControl::Build() {
   }
 
   // Construct QP
-  n_h_ = distance_constraints_.size();
-  n_c_ = 3 * body_indices_.size();
   prog_ = std::make_unique<MathematicalProgram>();
 
+  // Size of decision variable
+  n_h_ = (kinematic_evaluators_ == nullptr)
+             ? 0
```
yangwill, line 248: formatting looks weird here, I would think `? 0 : kinematic_evaluators_->count_full()` would be on the same line

systems/controllers/osc/operational_space_control.cc
```diff
@@ -381,8 +393,18 @@ VectorXd OperationalSpaceControl::SolveQp(
     const VectorXd& x_w_spr, const VectorXd& x_wo_spr,
     const drake::systems::Context<double>& context, double t, int fsm_state,
     double time_since_last_state_switch) const {
-  vector<bool> active_contact_flags = CalcActiveContactIndices(fsm_state);
+  // Get active contact indices
+  vector<int> active_contact_list = {};
+  if (single_contact_mode_) {
+    active_contact_list = contact_indices_map_.at(-1);
+  } else {
+    auto it_map = contact_indices_map_.find(fsm_state);
```
yangwill, line 401: we should either be doing something if the fsm_state isn't found, or just not checking for the fsm_state here and doing the check elsewhere.

systems/controllers/osc/operational_space_control.cc
```diff
@@ -381,8 +393,18 @@ VectorXd OperationalSpaceControl::SolveQp(
     const VectorXd& x_w_spr, const VectorXd& x_wo_spr,
     const drake::systems::Context<double>& context, double t, int fsm_state,
     double time_since_last_state_switch) const {
-  vector<bool> active_contact_flags = CalcActiveContactIndices(fsm_state);
+  // Get active contact indices
+  vector<int> active_contact_list = {};
```
yangwill, line 397: Because we only use this to check if a contact index exists in the active set, this really should be a `std::set<int>` instead of a `std::vector<int>`

systems/controllers/osc/operational_space_control.cc
```diff
@@ -396,33 +418,42 @@ VectorXd OperationalSpaceControl::SolveQp(
...
-          plant_wo_spr_
-              .CalcBiasForJacobianSpatialVelocity(
-                  *context_wo_spr_, JacobianWrtVariable::kV,
-                  plant_wo_spr_.get_body(body_indices_[i]).body_frame(),
-                  pts_on_body_[i], world_wo_spr_, world_wo_spr_)
-              .tail(3);
+  for (unsigned int i = 0; i < all_contacts_.size(); i++) {
+    if (std::find(active_contact_list.begin(), active_contact_list.end(), i) !=
+        active_contact_list.end()) {
+      J_c.block(3 * i, 0, 3, n_v_) =
```
yangwill, line 434: although it may be obvious, we should use a constant instead of the number 3. Maybe name the constant `static const int SPACE_DIM` or something similar

systems/controllers/osc/operational_space_control.cc
```diff
@@ -396,33 +418,42 @@ VectorXd OperationalSpaceControl::SolveQp(
...
+  int row_idx = 0;
+  for (unsigned int i = 0; i < all_contacts_.size(); i++) {
+    auto contact_i = all_contacts_[i];
+    if (std::find(active_contact_list.begin(), active_contact_list.end(), i) !=
+        active_contact_list.end()) {
+      // We don't call EvalActiveJacobian() because it'll repeat the computation
+      // of the Jacobian. (J_c_active is just a stack of slices of J_c)
+      for (int j = 0; j < contact_i->num_active(); j++) {
+        J_c_active.row(row_idx + j) =
+            J_c.row(3 * i + contact_i->active_inds().at(j));
```
yangwill, line 451: the constant would be helpful here as well

systems/controllers/osc/operational_space_control.cc
```diff
@@ -248,67 +271,68 @@ void OperationalSpaceControl::Build() {
...
+                   VectorXd::Zero(n_c_active_), {dv_, epsilon_})
+               .evaluator()
+               .get();
     }
   }
-  // 4. Friction constraint (approximated firction cone)
-  if (!body_indices_.empty()) {
+  // 4. Friction constraint (approximated friction cone)
+  if (!all_contacts_.empty()) {
     VectorXd mu_minus1(2);
```
yangwill, line 297: we should rename `mu_minus1` and  `mu_plus1` to `mu_neg1` and `mu_1` because `mu_minus1` sounds like it is `(mu - 1)` instead of `mu, -1`

systems/controllers/osc/operational_space_control.h
```diff
@@ -250,22 +250,16 @@ class OperationalSpaceControl : public drake::systems::LeafSystem<double> {
...
-  // constraint is always active.
-  // The states here can repeat, since there might be multiple contact points
-  // in a state of the finite state machine.
-  std::vector<int> fsm_state_when_active_;
-
-  // CalcActiveContactIndices gives a vector of flags indicating the active
-  // contact (constraint)
-  std::vector<bool> CalcActiveContactIndices(int fsm_state) const;
+  // Map finite state machine state to its active contact indices
+  std::map<int, std::vector<int>> contact_indices_map_ = {};
```
yangwill, line 258: I think this would be more useful as a `std::map<int, std::set<int>>`

systems/controllers/cp_traj_gen.cc
```diff
@@ -156,28 +147,16 @@ void CPTrajGenerator::calcCpAndStanceFootHeight(
...
-  } else {
-    stance_foot_idx = left_right_foot_body_indices_[0];
-    pt_on_stance_foot = left_right_foot_pts_on_bodies_[0];
-  }
-  Vector3d stance_foot_pos =
-      tree_.transformPoints(cache, pt_on_stance_foot, stance_foot_idx, 0);
+  plant_.SetPositions(context_.get(), q);
+
+  // Stance foot position
+  auto stance_foot = (fsm_state(0) == left_right_support_fsm_states_[1])
```
yangwill, line 154: this should be a map from the fsm_state to the stance_foot

examples/Cassie/run_osc_standing_controller.cc
yminchen: The loop closure flag in the function creates spring/damper element to enforce the distance constraint, but I think what we need in OSC is a kinematic distance constraint

examples/Cassie/run_osc_standing_controller.cc
yminchen: Done

systems/controllers/osc/operational_space_control.cc
yminchen: I renamed it to `map_iterator`

systems/controllers/osc/operational_space_control.cc
yminchen: I'm not sure what the correct format is, but I changed it to
```
  n_h_ = (kinematic_evaluators_ == nullptr)?
      0 : kinematic_evaluators_->count_full();
```

systems/controllers/osc/operational_space_control.cc
yminchen: Correct, I didn't realize what I needed was `std::set`

systems/controllers/osc/operational_space_control.cc
yminchen: Done

systems/controllers/osc/operational_space_control.cc
yminchen: Done

systems/controllers/osc/operational_space_control.cc
yminchen: done

systems/controllers/cp_traj_gen.cc
yminchen: Done

systems/controllers/osc/operational_space_control.cc
yminchen: Currently, if the `fsm_state` is not found in the map, it uses an empty contact set. What do you think we should do in this case?

yangwill: An empty contact set should work, we should check to make sure the fsm_state is valid though. It might lead to hard to debug errors further later if we just assume there are no active contacts if we can't find the fsm_state.

yminchen: Ok, I added a little print message so that it's easy to spot the error. If we find a better solution later, we can update this

PR #164, opened 2020-06-18
Port Cassie's dispatchers and related classes

yminchen:
Changes: 
* Port `CassieStateEstimator` (also clean up codes and fix a few minor bugs) 
* Port `dispatcher_robot_in`, `dispatcher_robot_out`, and `input_supervisor`
* Rename `LeftHeel()` and `RightHeel()` in `cassie_utils` to avoid confusion with the achilles rods on heels

Cassie is able to walk and stand with EKF in the simulation


examples/Cassie/dispatcher_robot_out.cc
```diff
@@ -2,69 +2,80 @@
...
                         drake::systems::Context<double>& diagram_context,
                         double t0) {
   auto& state_estimator_context =
       diagram.GetMutableSubsystemContext(*state_estimator, &diagram_context);
   state_estimator->setPreviousTime(&state_estimator_context, t0);
-  state_estimator->setInitialImuPosition(
-      &state_estimator_context, Eigen::Vector3d(0.0318638, 0, FLAGS_init_imu_height));
-  state_estimator->setInitialImuQuaternion(&state_estimator_context,
-                                           Eigen::Vector4d(1, 0, 0, 0));
+  state_estimator->setInitialPevlisPose(
```
yangwill, line 71: Typo: should be setInitialPelvisPose()

examples/Cassie/cassie_state_estimator.h
```diff
@@ -71,12 +73,11 @@ class CassieRbtStateEstimator : public drake::systems::LeafSystem<double> {
       int* left_contact, int* right_contact) const;
 
   void setPreviousTime(drake::systems::Context<double>* context, double time);
-  void setInitialImuPosition(drake::systems::Context<double>* context,
-                             Eigen::Vector3d p);
-  void setInitialImuQuaternion(drake::systems::Context<double>* context,
-                               Eigen::Vector4d q);
+  void setInitialPevlisPose(drake::systems::Context<double>* context,
```
yangwill, line 76: Typo in Pelvis

examples/Cassie/cassie_state_estimator.cc
```diff
@@ -232,60 +263,60 @@ CassieRbtStateEstimator::CassieRbtStateEstimator(
...
         r_ball_joint - r_heel_spring_base;
     Vector3d r_thigh_ball_joint_wrt_heel_spring_base =
-        heel_spring_rot_mat.transpose() * r_heel_spring_base_to_thigh_ball_joint;
+        heel_spring_rot_mat.transpose() *
+        r_heel_spring_base_to_thigh_ball_joint;
 
     // Get the projected rod length in the xy plane of heel spring base
-    double projected_rod_length = sqrt(pow(rod_length_, 2) -
-        pow(r_thigh_ball_joint_wrt_heel_spring_base(2), 2));
+    double projected_rod_length =
```
yangwill, line 298: let's not use `pow()` to compute squares, I think `rod_length_ * rod_length_` would be easier to read.

examples/Cassie/cassie_state_estimator.cc
```diff
@@ -1312,40 +1231,51 @@ EventStatus CassieRbtStateEstimator::Update(const Context<double>& context,
...
-  const auto& cassie_out = this->EvalAbstractInput(
-      context, cassie_out_input_port_)->get_value<cassie_out_t>();
+void CassieStateEstimator::CopyStateOut(const Context<double>& context,
+                                        OutputVector<double>* output) const {
+  const auto& cassie_out =
+      this->EvalAbstractInput(context, cassie_out_input_port_)
+          ->get_value<cassie_out_t>();
   // There might be a better way to initialize?
-  auto data = output->get_mutable_data();  // This doesn't affect timestamp value
+  auto data =
```
yangwill, line 1277: what is the purpose of zero-ing out the output vector?

yangwill:
Looks mostly good, though we should probably resolve some of the TODOs. Let me know if you want my help with that.

examples/Cassie/dispatcher_robot_out.cc
yminchen: done

examples/Cassie/cassie_state_estimator.h
yminchen: done

examples/Cassie/cassie_state_estimator.cc
yminchen: I forgot why we added this initialization. I'm guessing we didn't have imu data back then, and having some values being nan was causing trouble. 

I am taking this off, since the current code seems working fine without it.

yminchen:
@yangwill I removed the finished TODOs, and the rest of the TODOs in state estimators has to wait for hardware experiment

examples/Cassie/cassie_utils.h
```diff
@@ -51,21 +51,39 @@ void addCassieMultibody(drake::multibody::MultibodyPlant<double>* plant,
     std::string filename = "examples/Cassie/urdf/cassie_v2.urdf",
     bool add_leaf_springs = true,  bool add_loop_closure = true);
 
+double AchillesLength();
```
yangwill, line 54: I think the best way to return a constant double is to list it as a `static constexpr double` member variable. Might be useful making it public just in case we need to call it again.

yminchen: done

mposa: Can you give it a better name? https://google.github.io/styleguide/cppguide.html#Constant_Names

yminchen: Ah yes. Thanks for the reference!

PR #167, opened 2020-06-26
Connect osc debug port for the walking controller

yminchen:
I connected the port so that `OscTrackingData`'s are published through lcm and Joan can potentially use this as an example for his visualization work


examples/Cassie/run_osc_walking_controller.cc
```diff
@@ -117,6 +119,11 @@ int DoMain(int argc, char* argv[]) {
   builder.Connect(command_sender->get_output_port(0),
                   command_pub->get_input_port());
 
+  // Create osc debug sender.
+  auto osc_debug_pub =
```
mposa, line 123: Only create the system if it's going to be connected (if the flag is set). Can probably do both in the same if statement.

yminchen: done

PR #169, opened 2020-07-01
Fix a bug and improve dircon squatting script 

yminchen:
Fix scaling bugs, rename a variable, use bounding box constraint, scale the cost and provide initial guesses for slack/time variables

### DIFF
--- a/examples/Cassie/run_dircon_squatting.cc
+++ b/examples/Cassie/run_dircon_squatting.cc
@@ -190,10 +190,11 @@ void DoMain(double duration, int max_iter, string data_directory,
     // Kinematic constraints
     int n_kin = double_all_dataset.countConstraints();
     double s_kin_vel = 500;
-    double s_kin_acc = 1000;
+    double s_kin_pos = 1000;
     double s_kin_1 = (FLAGS_is_scale_variable) ? 10.0 : 1.0;
     double s_kin_2 = (FLAGS_is_scale_variable) ? 2.0 : 1.0;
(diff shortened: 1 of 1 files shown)


yminchen:
> What's the effect on benchmark runtime, if any?

Here is what I observed about the solve time:
* Fixing the constraint scaling bug: solve time barely changed
* Using bounding box constraints: slightly decreased
* Scaling the cost: increased by ~50%
* Providing initial guesses for the slack and time variables: decreased by 50~100%

I am not sure why scaling the cost didn't help to improve the solve time in this example

mposa:
> * Scaling the cost: increased by ~50%
So why scale the cost in this PR?

> * Providing initial guesses for the slack and time variables: decreased by 50~100%
100% reduction = instantaneous?

yminchen:
Ah I made some mistakes in the previous comment. I benchmarked solve time more carefully this time and here is the result: 
* Fixing the constraint scaling bug: increased by ~60 seconds
* Using bounding box constraints: decreased slightly
* Providing initial guesses for the slack and time variables: decreased by ~10 seconds
* Scaling the cost: reduced by ~50 seconds
The above changes is listed in order. With a different order, we might get a different solve time result.

Interestingly, the unintentional bug increased the solving speed

mposa:
1/500 is a pretty big scale factor to include if it doesn't actually improve performance. Might be worth dialing it back to see what happens.

yminchen:
Ok, I'll close this PR for now

PR #172, opened 2020-07-02
DIRCON rework.

mposa:
Rewrites DIRCON code to use new evaluators and an impoved API. In addition, contains two related changes:

-Add friction constraints as options to kinematic evaluators
-Change multibody_utils CreateContext and ModifyContext to take Eigen::Ref, avoiding a small copy operation. Note that this requires manual template declarations, e.g. `CreateContext<double>(...)` as template type cannot be inferred automatically.

This PR introduces one new example, and rewrites two existing examples. **It does not delete the old code or rewrite all examples**, as this should be the product of a later PR once we are happy with the rewrite performance. The examples are:
-(new) systems/trajectory_optimization/dircon/test/passive_constrained_pendulum_dircon.cc: A floating-base acrobot constrained via a WorldPointEvaluator and DistanceEvaluator to behave like a passive pendulum. Tests both evaluator types and quaternion constraints.
-(updated) examples/PlanarWalker/run_gait_dircon.cc 
-(updated) examples/Cassie/run_dircon_squatting.cc

Feedback on the code, API, and examples is all appreciated.

## Caching
I experimented with a more aggressive Systems-based caching mechanism using individual dependencies on `q,v,lambda`, but eventually abandoned it in favor of a more narrowly scoped DIRCON-specific cache for `xdot` calculations only. The more aggressive cache was going to be very difficult to maintain, and essentially required manual determination of sparsity patterns and gradient extraction/creation. 

While there's some room for improvement, `xdot` is definitely the most expensive repeated computation. What this fails to capture, however, is that not every element used in `xdot` requires differentiation w.r.t. all inputs `x,u,lambda`. For instance, the bias terms are independent of `u` and `lambda`, and the mass matrix only depends on `q`. Profiling shows that future improvements to numerical gradient calculations could show benefits.

## Scaling
I noted, as everyone has before me, extreme sensitivity on runtime w.r.t. scaling parameters on the Cassie example, while using SNOPT. I also tried using IPOPT, where the default automatic scaling seemed to render our manual scaling unnecessary. While SNOPT with optimized-scale parameters is ~2x faster, on limited testing I found IPOPT to be more reliable for the squatting example. Note that the example now includes an option to use IPOPT, along with some option settings. For these options, IPOPT did not converge to dual feasibility, but was primal feasible and had stalled progress on improving the cost (achieving the same cost as SNOPT).

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/172)


mposa:
I also found a bug in the old Dircon code, where certain forces were not being included in the dynamics calculations. Specifically, the old code ignored joint damping. Hopefully resolving https://github.com/RobotLocomotion/drake/issues/13476 will make this easier in the future.

multibody/kinematic/distance_evaluator.cc
```diff
@@ -78,37 +90,38 @@ VectorX<T> DistanceEvaluator<T>::EvalFullJacobianDotTimesV(
...
-      pt_A_cast, world, world).translational();
-  VectorX<T> J_B_dot_times_v = plant().CalcBiasSpatialAcceleration(
-      context, drake::multibody::JacobianWrtVariable::kV, frame_B_,
-      pt_B_cast, world, world).translational();
+  VectorX<T> J_A_dot_times_v =
+      plant()
+          .CalcBiasSpatialAcceleration(
+              context, drake::multibody::JacobianWrtVariable::kV, frame_A_,
+              pt_A_cast, world, world)
+          .translational();
```
yangwill, line 115: I think we should be able to use `CalcBiasTranslationalAcceleration()` in place of `CalcBiasSpatialAcceleration().translational()`

mposa: Good to know! I missed that one. Fixed in WorldPointEvaluator as well.

multibody/kinematic/kinematic_evaluator.h
```diff
@@ -84,6 +85,28 @@ class KinematicEvaluator {
...
+  const std::vector<int>& active_inds() { return active_inds_; };
+
+  bool is_active(int index) const;
+
+  /// Convert an index, relative to the full index set, to be an active index.
+  /// Returns -1 if the given index is not active
+  int full_index_to_active_index(int full_index) const;
+
+  /// Create a friction cone constraint on the force variables (associated with
+  /// the full Jacboian). Subclasses which might be associated with frictional
```
yangwill, line 97: typo in "Jacobian"

multibody/kinematic/kinematic_evaluator_set.cc
```diff
@@ -271,6 +305,21 @@ int KinematicEvaluatorSet<T>::count_full() const {
...
 }
 
+template <typename T>
+bool KinematicEvaluatorSet<T>::is_active(int index) const {
+  if (index < 0 || index >= count_full()) {
+    return false;
+  }
+  int count = 0;
+  for (const auto& e : evaluators_) {
+    if (index - count >= 0 && index - count < e->num_full()) {
```
yangwill, line 315: This block took a minute to understand - is it worth creating a separate vector that stores whether each line of all the evaluators are active? The only use case I can think of for creating such a vector is to be able to print out which constraints are active, thoughts?

systems/trajectory_optimization/dircon/dircon.cc
```diff
@@ -0,0 +1,843 @@
...
+    // Create and add kinematic constraints
+    //
+    for (int j = 0; j < mode.num_knotpoints(); j++) {
+      // Position constraints if type is All
+      if (mode.get_constraint_type(j) == KinematicConstraintType::kAll) {
+        VectorXd lb = VectorXd::Zero(mode.evaluators().count_active());
+        VectorXd ub = VectorXd::Zero(mode.evaluators().count_active());
+
+        // If we are in the first knotpoint, do not constrain the position of
+        // any constraints that overlap with the o avoirevious mode, as such a
```
yangwill, line 162: Typo? seems like a bad find/replace

systems/trajectory_optimization/dircon/dircon.h
```diff
@@ -0,0 +1,221 @@
...
+      const drake::solvers::MathematicalProgramResult& result) const override;
+
+  /// Get the state trajectory at the solution as a
+  /// %drake::trajectories::PiecewisePolynomialTrajectory%.
+  drake::trajectories::PiecewisePolynomial<double> ReconstructStateTrajectory(
+      const drake::solvers::MathematicalProgramResult& result) const override;
+ 
+  /// Get the state samples by mode, as a matrix. Each column corresponds to
+  /// a knotpoint.
+  Eigen::MatrixXd GetStateSamplesByMode(
```
yangwill, line 55: Cool this looks like exactly what I wanted!

yangwill:
Only made if about halfway through the review, will continue tomorrow.

systems/trajectory_optimization/dircon/dircon.cc
```diff
@@ -0,0 +1,843 @@
...
+          break;
+        }
+      }
+    }
+
+    //
+    // Create and add collocation constraints
+    //
+    int cache_size = 1000;
+    cache_.push_back(
```
yangwill, line 135: Have you tested different cache sizes? Does 1000 seem to be the sweet spot? If so, maybe make it a constant? If not, add an option to adjust it.

systems/trajectory_optimization/dircon/dircon_mode.h
```diff
@@ -0,0 +1,231 @@
...
+  };
+
+  /// Declare that quaternion constraints at the given knotpoint_index are not
+  /// to be included. This is useful if the quaternion configuration is already
+  /// constrained via another mechanism.
+  void SkipQuaternion(int knotpoint_index);
+
+  bool IsSkipQuaternion(int knotpoint_index) const;
+
+  /// Count the nubmer of relative constraints
```
yangwill, line 58: typo

systems/trajectory_optimization/dircon/dircon_mode.h
```diff
@@ -0,0 +1,231 @@
...
+
+  /// Gets the map of non-unitary scales for the kinematic acceleration
+  /// constraints. Note, unlike GetKinAccelerationScale, indices in the map
+  /// refer to the complete KinematicEvaluatorSet, not a specific
+  /// KinematicEvaluator.
+  const std::unordered_map<int, double>& GetKinAccelerationScale() const {
+    return kin_acceleration_scale_;
+  };
+
+  /// Get the cost on force regularization. Defaults to 1e-4
```
yangwill, line 162: 1e-4 doesn't match the default value used in the constructor

systems/trajectory_optimization/dircon/dircon_mode.h
```diff
@@ -0,0 +1,231 @@
...
+
+  const multibody::KinematicEvaluatorSet<T>& evaluators_;
+  const int num_knotpoints_;
+  const double min_T_;
+  const double max_T_;
+  const double force_regularization_;
+  double mu_;
+  std::set<int> relative_constraints_;
+  std::set<int> skip_quaternion_;
+  std::unordered_map<int, KinematicConstraintType> reduced_constraints_;
```
yangwill, line 195: It's not immediately clear what this is from the variable name.

examples/Cassie/run_dircon_squatting.cc
```diff
@@ -106,176 +109,178 @@ void DoMain(double duration, int max_iter, string data_directory,
...
-  double_all_options.setConstraintRelative(1, true);
-  double_all_options.setConstraintRelative(3, true);
-  double_all_options.setConstraintRelative(5, true);
-  double_all_options.setConstraintRelative(6, true);
-  double_all_options.setConstraintRelative(8, true);
+  auto double_support =
+      DirconMode<double>(evaluators, num_knotpoints, min_T, max_T);
+  double_support.set_mu(mu);
+  // Set x-y coordinates as relative
+  double_support.MakeConstraintRelative(0, 0);  // left_toe, x
```
yangwill, line 160: In case someone wants to use this as a reference for creating a DIRCON problem, it might be useful to use the index returned from `evaluators.add_evaluator(&left_toe_eval)` for the first argument.

multibody/kinematic/kinematic_evaluator_set.cc
mposa: Like you said, I think that's the only real use case. I'd worry a little about ensuring that the vector stays synchronized. For instance, it could not be a property of the `Set` since the individual `Evaluators` can be updated without the `Set` knowing. I added a comment to explain what the code is doing. Let me know if that helps.

systems/trajectory_optimization/dircon/dircon.cc
mposa: I'll set it intelligently. A safe number is equal to the number of decision variables (plus 1) for numerical gradients.

examples/Cassie/run_dircon_squatting.cc
mposa: Good idea.

multibody/kinematic/kinematic_evaluator.h
```diff
@@ -72,17 +76,35 @@ class KinematicEvaluator {
...
+  int full_index_to_active_index(int full_index) const;
 
-  int num_active() const {
-    return num_active_;
-  }
+  /// Create a friction cone constraint on the force variables (associated with
+  /// the full Jacobian). Subclasses which might be associated with frictional
+  /// contact should implement this method.
+  virtual std::shared_ptr<drake::solvers::Constraint>
+  CreateConicFrictionConstraint(double mu) const {
```
yminchen, line 97: Is there going to be another evaluator that has friction constraint? If not, should we move the friction constraint methods to WorldPointEvaluator?

yminchen:
I was hoping to find a clean-ish implementation of a caching system for our numerical differentiation. And not surprising, it doesn't look clean... (especially because dircon is templated) Below is what I came up with. 

I think we are only using Drake's context to cache forward kinematics elements, J and JdotV, so the robot's state `x` is the only cache dependency here. Assuming this is true, we can create `n_x + 1` number of `Context` for each knot point. 
Dircon constructor could look like:
```
// Loop over all modes
for (int i_mode = 0; i_mode < num_modes(); i_mode++) {
  ...
 
  //
  // Create context elements 
  //
  if (typeid(T) == typeid(drake::AutoDiffXd)) {
    // create 1 context per knot point
  } else if (typeid(T) == typeid(double)) {
    // create n_x + 1 context per knot point
  } else {
    DRAKE_UNREACHABLE();
  }
  
  ...
}

```
To handle the cache dependency for numerical differentiation, I think we can create a class derived from `NonlinearConstraint`, and we override the `DoEval` where we compute gradient numerically. The new constraint class takes in `dependent_input_inds` which tells the location of `x` (or `x`'s) in the constraint input. And we add a virtual method `EvaluateConstraintWithCache` which is similar to `EvaluateConstraint` except for an extra input which tells the derived class which `context` it should use among `n_x + 1` contexts.  
```
NonlinearConstraintWithCache<T>::NonlinearConstraintWithCache(
    int num_constraints, int num_vars, const VectorXd& lb, const VectorXd& ub,
    const std::set<int> dependent_input_inds, const std::string& description)
    : NonlinearConstraint(num_constraints, num_vars, lb, ub, description) {
  // DRAKE_DEMAND each element of dependent_input_inds < num_vars

  // Create the complement of dependent_input_inds
}

// NonlinearConstraintWithCache has a pure virtual function
// EvaluateConstraintWithCache(const Eigen::Ref<const drake::VectorX<T>>& x,
//                             int cache_id, drake::VectorX<T>* y)

template <>
void NonlinearConstraintWithCache<double>::DoEval(
    const Eigen::Ref<const AutoDiffVecXd>& x, AutoDiffVecXd* y) const {
  MatrixXd original_grad = drake::math::autoDiffToGradientMatrix(x);

  // forward differencing
  VectorXd x_val = drake::math::autoDiffToValueMatrix(x);
  VectorXd y0, yi;
  EvaluateConstraintWithCache(x_val, 0, &y0);

  MatrixXd dy = MatrixXd(y0.size(), x_val.size());
  for (int idx = 0; idx < dependent_input_inds.size(); idx++) {
    x_val(dependent_input_inds(idx)) += eps_;
    EvaluateConstraintWithCache(x_val, idx, &yi);
    x_val(dependent_input_inds(idx)) -= eps_;
    dy.col(dependent_input_inds(idx)) = (yi - y0) / eps_;
  }
  for (int idx = 0; idx < complement_of_dependent_input_inds.size(); idx++) {
    x_val(complement_of_dependent_input_inds(idx)) += eps_;
    EvaluateConstraintWithCache(x_val, 0, &yi);
    x_val(complement_of_dependent_input_inds(idx)) -= eps_;
    dy.col(complement_of_dependent_input_inds(idx)) = (yi - y0) / eps_;
  }

  // Profiling identified dy * original_grad as a significant runtime event,
  // even though it is almost always the identity matrix.
  if (original_grad.isIdentity(1e-16)) {
    *y = drake::math::initializeAutoDiffGivenGradientMatrix(y0, dy);
  } else {
    *y = drake::math::initializeAutoDiffGivenGradientMatrix(y0,
                                                            dy * original_grad);
  }

  this->ScaleConstraint<AutoDiffXd>(y);
}

```   
We can pass `std::vector<Context>` into `DirconCollocationConstraint`, `ImpactConstraint` and `CachedAccelerationConstraint` to avoid calling two versions of constructor for double and autodiff. 
Haven't thought through all the details but just to present an idea here

yminchen:
> _[multibody/kinematic/kinematic_evaluator_set.cc, line 105 at r5](https://reviewable.io/reviews/dairlab/dairlib/172#-MCd8pH1cXoqbutjcezG-r5-105:-MCdTdel-1Am-Uf7gTlD:b-2b36sx) ([raw file](https://github.com/dairlab/dairlib/blob/6b8aa03646d6bfc497cdb6a9250fd4f1bc5fdf63/multibody/kinematic/kinematic_evaluator_set.cc#L105)):_
> Previously, yminchen (Yu-Ming Chen) wrote…
> 
> No, it should be `vdot` from applying chain rule to `phi(q)`
> 
It looks to me that `KinematicEvaluatorSet<T>::CalcTimeDerivativesWithForce` always returns `xdot`. https://github.com/DAIRLab/dairlib/blob/9c8e722/multibody/kinematic/kinematic_evaluator_set.cc#L244
Am I not looking at the correct function? 



> _[multibody/kinematic/world_point_evaluator.cc, line 90 at r5](https://reviewable.io/reviews/dairlab/dairlib/172#-MCd8pH1cXoqbutjcezJ-r5-90:-MCdXM6I0yK9EjR3ROv9:b-sro69l) ([raw file](https://github.com/dairlab/dairlib/blob/6b8aa03646d6bfc497cdb6a9250fd4f1bc5fdf63/multibody/kinematic/world_point_evaluator.cc#L90)):_
> Previously, yminchen (Yu-Ming Chen) wrote…
> 
> It's not included as a bounding box constraint, but implied by the single nonlinear constraint.
> 
> Should we go back to something like the current formulation where `Evaluator.CreateXXFrictionConeConstraint` actually returns a list of constraints?
> 
Yea, returning a list of constraints should be a good solution for now


> 
> _[systems/trajectory_optimization/dircon/dircon.h, line 214 at r5](https://reviewable.io/reviews/dairlab/dairlib/172#-MCd8pH28o2N58JfUNBv-r5-214:-MCdZ-Rw2PHKfNZNmxns:b-7rzf5y) ([raw file](https://github.com/dairlab/dairlib/blob/6b8aa03646d6bfc497cdb6a9250fd4f1bc5fdf63/systems/trajectory_optimization/dircon/dircon.h#L214)):_
> Previously, yminchen (Yu-Ming Chen) wrote…
> 
> Done. I added a getter to a `const Context<T>&`. Does that work? Exposing the pointer seems pointless, unless you see a reason.
> 
Yes that works!

> 
> _[systems/trajectory_optimization/dircon/dircon.cc, line 171 at r5](https://reviewable.io/reviews/dairlab/dairlib/172#-MCd8pH28o2N58JfUNBu-r5-171:-MCdZeMJ2AUezezTo51D:bbkha8z) ([raw file](https://github.com/dairlab/dairlib/blob/6b8aa03646d6bfc497cdb6a9250fd4f1bc5fdf63/systems/trajectory_optimization/dircon/dircon.cc#L171)):_
> Previously, yminchen (Yu-Ming Chen) wrote…
> 
> See if you think the fix to FindActiveUnion resolved this.
> 
The code looks correct, but the variable name should be `row_indices` instead of `union_indices`

multibody/kinematic/kinematic_evaluator_set.cc
```diff
@@ -117,22 +152,54 @@ template <typename T>
...
+    KinematicEvaluatorSet<T> other) const {
+  std::vector<int> row_indices;
   for (int i = 0; i < other.num_evaluators(); i++) {
     if (std::find(evaluators_.begin(), evaluators_.end(),
-        other.get_evaluator(i)) != evaluators_.end()) {
-      union_index.push_back(i);
+                  &other.get_evaluator(i)) != evaluators_.end()) {
+      int start_index = other.evaluator_active_start(i);
+      for (int j = 0; j < other.get_evaluator(i).num_active(); j++) {
+        row_indices.push_back(start_index + j);
```
yminchen, line 198: Returning indices of the full constraint might be better, and I think it's consistent to the rest of the APIs. 

The code probably looks like this: 
```
      int start_index = other.evaluator_full_start(i);
      for (int j = 0; j < other.get_evaluator(i).num_active(); j++) {
        row_indices.push_back(start_index + other.get_evaluator(i).active_inds().at(j));
      }
```

mposa:
Going to merge--but feel free to followup with other comments down the line as you use it.

PR #173, opened 2020-07-03
Goldilocks test

Jianshu-Hu:
Fix the bug in `initial_guess_test.cc`.

### DIFF
--- a/examples/goldilocks_models/find_models/initial_guess.h
+++ b/examples/goldilocks_models/find_models/initial_guess.h
@@ -16,7 +16,8 @@ std::string SetInitialGuessByInterpolation(const std::string& directory,
                                            int iter, int sample,
                                            const TasksGenerator* task_gen,
                                            const Task& task, const RomData& rom,
-                                           bool use_database, int robot);
+                                           bool use_database, int robot,
+                                           bool is_test=false);
 // set scale for theta and gamma
 Eigen::VectorXd GetThetaScale(const std::string& directory, int iter);
 Eigen::VectorXd GetGammaScale(const TasksGenerator* task_gen);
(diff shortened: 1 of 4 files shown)


examples/goldilocks_models/find_models/initial_guess_test.cc
```diff
@@ -32,16 +32,26 @@ int test_initial_guess(int iter, int sample, int robot) {
 
   const string dir =
       "../dairlib_data/goldilocks_models/find_models/robot_1_test/";
-  if (!CreateFolderIfNotExist(dir, false)) return 0;
```
yminchen, line 35: Why not use `CreateFolderIfNotExist`?

examples/goldilocks_models/find_models/initial_guess.h
```diff
@@ -16,7 +16,8 @@ std::string SetInitialGuessByInterpolation(const std::string& directory,
                                            int iter, int sample,
                                            const TasksGenerator* task_gen,
                                            const Task& task, const RomData& rom,
-                                           bool use_database, int robot);
+                                           bool use_database, int robot,
+                                           bool is_test=false);
```
yminchen, line 20: Since we can choose the value for `theta` in the test file, we shouldn't add a flag `is_test` here

examples/goldilocks_models/find_models/initial_guess_test.cc
Jianshu-Hu: Because I directly extract the parts used to create the folder from the `CreateFolderIfNotExist` and use it to create folder if it doesn't exist.

Jianshu-Hu:
1. Deleted the flag `is_test` and change the value of `theta` in test file.
2. As for the bugs, two main bugs: directory used to save temporary test data should be created successfully; the data file names should be consistent with the `initial_guess.cc`.

yminchen:
Maybe we can modify the code so that the expansion only happens in iteration 1. The only thing that we need to handle is probably the indexing. Maybe create a function that handles this during the expansion and a few iterations after the expansion?

PR #176, opened 2020-07-03
Small code cleanup

mposa:
-Move file_utils to /common
-Delete a bunch of unused files

@yminchen Can you confirm that you're not using any of the deleted files in your goldilocks branch?


yminchen:
> @yminchen Can you confirm that you're not using any of the deleted files in your goldilocks branch?

Nope, my branch doesn't use those deleted code

PR #178, opened 2020-07-05
Remove RigidBodyTree code from build

mposa:
Marks all remaining drake/attic code (RigidBodyTree) to only be built manually, and adds a deprecation warning. This code is not guaranteed to build anymore, but we can leave it for a few months until we're satisfied that it is not needed. Of course, it still will exist in git's history.

A number of lightly (or unused) programs fall into this category, including some of the KUKA-specific code. I've updated where it was just a matter of dependency cleanup, but have not re-written any examples that are fully-based in RigidBodyTree.

This update also adds the IMU sensors to Cassie.

I tried both Cassie walking/balancing examples with the EKF, but both failed. The EKF in the balancing example was clearly not working. A very quick look did not show any _obvious_ problems with the IMU readings (near steady-state while balancing). @yminchen could you take a quick look at this branch and see if you can tell if the EKF is behaving properly, as I know this was recently updated?


yminchen:
@mposa Looks like the QPs for contact estimation are not giving good solutions. I checked the imu measurement is correct. I'm wondering if the equations of motion are not constructed correctly 
https://github.com/DAIRLab/dairlib/blob/master/examples/Cassie/cassie_state_estimator.cc#L522-L528

mposa:
> @mposa Looks like the QPs for contact estimation are not giving good solutions. I checked the imu measurement is correct. I'm wondering if the equations of motion are not constructed correctly
> https://github.com/DAIRLab/dairlib/blob/master/examples/Cassie/cassie_state_estimator.cc#L522-L528

Compare to this (which I'm pretty sure is correct for our examples, though potentially imperfect in how it includes generalized forces--there's an open issue in Drake to improve the API)
https://github.com/DAIRLab/dairlib/blob/e5d3ac0186e1c09c0bcafa487687331565dbba3f/multibody/kinematic/kinematic_evaluator_set.cc#L179-L193

yminchen:
I found the issue comes form the distance constraint of the QPs in EKF. Because the loop closure in the simulation is implemented via a spring, the displacement of this virtual spring is translated to the real ankle spring deflection in the QPs which leads to inaccurate dynamic prediction.

I'm guessing this virtual spring also affects the OSC solution, but I haven't looked into it yet.

yminchen:
Benchmark of the optimal QP cost in EKF:  
* The cost for the double support, including dynamics error and soft constraint error, is about 1e-5 when we use distance constraint in simulation. 
* The cost is about 250 when we use spring stiffness of 2e5 for loop closure
* The cost is about 20 when we use spring stiffness of 1e6 for loop closure

mposa:
> Benchmark of the optimal QP cost in EKF:
> 
> * The cost for the double support Cassie, including dynamics error and soft constraint error, is about 1e-5 when we use distance constraint in simulation.
> * The cost is about 250 when we use spring stiffness of 2e5 for loop closure in simulation
> * The cost is about 20 when we use spring stiffness of 1e6 for loop closure in simulation

Is the error in the spring constraint of a realistic scale, or is it unreasonably large? If the former, then we should be robust to it.

yminchen:
> Is the error in the spring constraint of a realistic scale, or is it unreasonably large? If the former, then we should be robust to it.

I looked at the value of the spring deflection for the loop closure in Drake simulation. With stiffness = 1e6, the deflection is about 0.0004 meter at steady state, which could probably be realistic? 

Roughly calculating the stiffness of the achilles rod with the equation given by https://physics.stackexchange.com/questions/320089/what-is-the-spring-constant-of-a-steel-wire,   
k = E*A/L = 1e10 * 1e-4 / 0.5 = 2e6
where I assume the young's modulus E is 1e10 pa (Cassie's should be higher? https://www.engineeringtoolbox.com/young-modulus-d_773.html), the area A is 1 cm^2, and length is 0.5 m.
I'm guessing simulated stiffness 1e-6 is an underestimate, but maybe it's not too far from the real value.

I will check the value of the QP cost is when Cassie stands still in experiment. This will tell us how close our model is to the robot, and I'll tune the EKF to be robust against the model error

mposa:
> > Is the error in the spring constraint of a realistic scale, or is it unreasonably large? If the former, then we should be robust to it.
> 
> I looked at the value of the spring deflection for the loop closure in Drake simulation. With stiffness = 1e6, the deflection is about 0.0004 meter at steady state, which could probably be realistic?
> 
> Roughly calculating the stiffness of the achilles rod with the equation given by https://physics.stackexchange.com/questions/320089/what-is-the-spring-constant-of-a-steel-wire,
> k = E*A/L = 1e10 * 1e-4 / 0.5 = 2e6
> where I assume the young's modulus E is 1e10 pa (Cassie's should be higher? https://www.engineeringtoolbox.com/young-modulus-d_773.html), the area A is 1 cm^2, and length is 0.5 m.
> I'm guessing simulated stiffness 1e-6 is an underestimate, but maybe it's not too far from the real value.
> 
> I will check the value of the QP cost is when Cassie stands still in experiment. This will tell us how close our model is to the robot, and I'll tune the EKF to be robust against the model error

Ok! Good sanity check on Young's Modulus, though I'd say that compliance in a robot typically comes from the inclusion of other more flexible elements (like bushings) or from beam bending rather than pure tension/compression.

yminchen:
> Ok! Good sanity check on Young's Modulus, though I'd say that compliance in a robot typically comes from the inclusion of other more flexible elements (like bushings) or from beam bending rather than pure tension/compression.

A correction on the Young's modulus -- I forgot to time 6.8 when converting psi to pa. The real stiffness is probably higher than 1e7

PR #182, opened 2020-07-23
Update Dockerfile

mposa:
A bug in InputSupervisor was caught by unit tests (hooray) combined with recent Drake updates. Some of our old code assumed a single discrete value (needed to use `[]` operator), but we had two declared. I updated the class to use a single, length 2, discrete value.

Also increments the Dockerfile version tag to trigger a rebuild with a fresh dependency installation.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/182)

### DIFF
--- a/examples/Cassie/input_supervisor.cc
+++ b/examples/Cassie/input_supervisor.cc
@@ -1,4 +1,5 @@
 #include "examples/Cassie/input_supervisor.h"
+
 #include "systems/framework/output_vector.h"
 
 using drake::systems::Context;
@@ -24,7 +25,8 @@ InputSupervisor::InputSupervisor(
   command_input_port_ =
       this->DeclareVectorInputPort(TimestampedVector<double>(num_actuators_))
           .get_index();
-  state_input_port_ = this->DeclareVectorInputPort(OutputVector<double>(
+  state_input_port_ = this
+                          ->DeclareVectorInputPort(OutputVector<double>(
                               num_positions_, num_velocities_, num_actuators_))
                           .get_index();
 
@@ -41,8 +43,10 @@ InputSupervisor::InputSupervisor(
           .get_index();
 
   // Create error flag as discrete state
-  n_consecutive_fails_index_ = DeclareDiscreteState(1);
-  status_index_ = DeclareDiscreteState(1);
+  // Store both values in single discrete vector
+  DeclareDiscreteState(2);
+  n_fails_index_ = 0;
+  status_index_ = 1;
 
   // Create update for error flag
   DeclarePeriodicDiscreteUpdateEvent(update_period, 0,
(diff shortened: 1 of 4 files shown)


examples/Cassie/input_supervisor.cc
```diff
@@ -55,7 +59,8 @@ void InputSupervisor::SetMotorTorques(const Context<double>& context,
       (TimestampedVector<double>*)this->EvalVectorInput(context,
                                                         command_input_port_);
 
-  bool is_error = context.get_discrete_state(0)[0] >= min_consecutive_failures_;
```
yangwill, line 58: I get that our previous hardcoding of the discrete state index made us prone to bugs. But what is the advantage of declaring a single discrete state vector of length 2 over declaring multiple discrete state vectors of length 1? Asking because a lot of our code uses the latter and I just want to be consistent.

The pros that I can see of using a single discrete state vector is that we can avoid the `[0]` indexing of a vector of length 1 which makes the code more readable.

Cons:
- we have to manually keep track of the indices to make sure there are no duplicates - not difficult but makes us slightly more error prone. 
- This indexing method fails when we need to declare a discrete state of length > 1.

mposa: I could see it either way. I'm happy to revert if you think it's better the other way. For what it's worth, this PR was triggered by Drake updating to catch bugs, and our CI catching that update. Previously, when we called `context.get_discrete_state()[i]` (or similar) it was secretly calling `get_discrete_state(0)[i]`. The Drake change is to only permit `context.get_discrete_state()[i]` when there is only one discrete state.

Basically, I think this approach makes sense for a set of length-1 discrete states. If an element of the state is, itself, a vector, then I don't think this is a good idea.

> The pros that I can see of using a single discrete state vector is that we can avoid the `[0]` indexing of a vector of length 1 which makes the code more readable.

Exactly, this is what I was thinking. I think it's more readable and easier to write this way for simple systems like this.

> Cons:
> * we have to manually keep track of the indices to make sure there are no duplicates - not difficult but makes us slightly more error prone.

What do you mean by duplicates? We need to track indices with either approach.

> * This indexing method fails when we need to declare a discrete state of length > 1.

Yes, true

yangwill: When you call `DeclareDiscreteState()` it returns a unique index. How it's implemented here, the user has to manually set each index. Again, not difficult to do, but opens us up to potential errors.

I think this PR is fine as is and for future code we can use this indexing scheme unless we are using >1 length discrete states. The majority of our code only uses length-1 discrete states anyway.

mposa: Good point.

PR #187, opened 2020-08-02
Add an input port in EKF to store lcm time

yminchen:
This PR fixes issue #63 

Instead of implementing the two solutions from Drake developers, I implemented a simpler approach here. I added an input port to EKF to store the message time. The port is updated after a new message arrives and before the simulation is advanced.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/187)

### DIFF
--- a/examples/Cassie/cassie_state_estimator.cc
+++ b/examples/Cassie/cassie_state_estimator.cc
@@ -70,10 +70,19 @@ CassieStateEstimator::CassieStateEstimator(
   n_v_ = plant.num_velocities();
   n_u_ = plant.num_actuators();
 
-  // Declare input/output ports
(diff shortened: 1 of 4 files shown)


mposa:
My opinion on this hasn't really changed since our earlier discussion. This feels _very wrong_ to me, since you're duplicating the notion of "time" which is already stored within the Context. The only problem with the current implementation is the ordering of how the update/step functions are called.

My preference would be option 1 from the suggested choices. I'm happy to write a rough framework for it tonight, but it'd probably be best if you were to test/finish the PR, since you better understand the EKF code.

yminchen:
Closing this PR, since we will probably use #188

PR #188, opened 2020-08-03
Change EKF update to registered event.

mposa:
Relates to #63 

Changes the EKF's update to be a kTimed event. The user must call `estimator.set_next_message_time(t)` to schedule an event at time `t` (technically, `t-eps`).

I'm not totally sure that this is the right approach, so please read with a critical eye. I'm also not totally sure it's better than #187 

Flaws:
- Since `DoCalcNextUpdateTime` is `const`, `next_message_time` is declared mutable so it can be reset within this method. This is probably not the right way to do this, as it breaks some of the Systems contracts regarding Context, but perhaps it's OK.
- The required use of `eps` in the event time. It could, alternatively, be moved to `simulator.AdvanceTo(t + eps)`, if that's any better.

The benefits from this approach are that:
- context.time is always correct (up to eps)
- Doesn't require superfluous states/ports to the system.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/188)

### DIFF
--- a/examples/Cassie/cassie_state_estimator.cc
+++ b/examples/Cassie/cassie_state_estimator.cc
@@ -943,320 +943,304 @@ EventStatus CassieStateEstimator::Update(
       this->EvalAbstractInput(context, cassie_out_input_port_)
           ->get_value<cassie_out_t>();
 
-  // TODO(yminchen): delete the testing code when you fix the time delay issue
-  // Testing
-//   cout << "\nIn per-step update: lcm_time = "
-//        << cassie_out.pelvis.targetPc.taskExecutionTime << endl;
-//   cout << "In per-step update: context_time = " << context.get_time() << endl;
-
   // Get current time and previous time
   double current_time = context.get_time();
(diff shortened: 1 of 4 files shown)


yminchen:
> Thanks for the quick fix here! Would it be better to avoid using `inf` and instead _not _ reset the time. Instead, in `DoCalcNextUpdateTime`, just check if
> 
> ```
> context.time < next_message_time
> ```
> 
> before declaring the event?
> 
Yeah I think that also works, with a slight modification, using `context.time < next_message_time - eps`. The changes are in the new commit

yminchen:
> Flaws:
> 
>     * Since `DoCalcNextUpdateTime` is `const`, `next_message_time` is declared mutable so it can be reset within this method. This is probably not the right way to do this, as it breaks some of the Systems contracts regarding Context, but perhaps it's OK.
`next_message_time` doesn't need to be mutable anymore in the new implementation!

PR #189, opened 2020-08-04
Sending high level commands using radio signal

yangwill:
Making a PR just to get feedback on implementation

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/189)


yminchen:
> Shouldn't this be taken care of by the driven loop?

I remember PerStep event could be triggered multiple times per lcm message, but I just checked Drake's code and tested our example, and it seems not to be the case now

PR #192, opened 2020-08-06
Add contact lcm

yminchen:
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/192)


yangwill:
Requesting input from @mposa. We are debating whether or not to include the contact state in `lcmt_robot_output`.
Some arguments for including it in `lcmt_robot_output`:
- We only have to pass a single lcm message for the robot's state
- The contact state is arguably a part of the robot's state

Some arguments against including it with robot output:
- The true contact forces from multibody_sim and mujoco are currently sent as a separate lcm message `drake::lcmt_contact_results_for_viz`
- Depending on how our contact estimation is implemented it might make more sense to update our estimates in a separate thread from the rest of the state estimation.
- Many places that use robot_output have no use for the contact state, though the same could be said for the imu_accel.

PR #193, opened 2020-08-07
Neutral point walking controller

yminchen:
Changes related to neutral point calculation only take about 3 lines of code. The rest are just renaming and setting options for both capture point controller and the new controller 

The *neutral point* is derived from LIPM dynamics. Given a stance duration and velocity at touchdown, we can compute the step location such that the center of mass exits the stance phase with the same velocity. 

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/193)


yminchen:
A few updates:
* Add a new leaf system `FiniteStateMachineEventTime` which outputs the latest switching time of the FSM
* Use the predicted COM velocity in the walking speed controller
* Fix the swing foot index bug we found in `cassie_experimental` branch

yminchen (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/193)* status: 14 of 16 files reviewed, 6 unresolved discussions (waiting on @yangwill)

___
*[examples/Cassie/osc/walking_speed_control.h, line 29 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGEKXuo7BTJVkPsET_C:-MKaPZy_EOAZ7i113hS6:b-htaqhl) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/examples/Cassie/osc/walking_speed_control.h#L29)):*

I think the k_d * (-rdot) is redundant if we have k_fb * (v - v_des). ie, rdot is equal to v. Though maybe it's more readable in this format?
</blockquote></details>

Yeah `rdot` is equal to `v`, but I think having them in two levels (velocity control and position control) makes it easier to tune the overall controller.
___
*[examples/Cassie/osc/walking_speed_control.cc, line 68 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGELOSq6FDpeKIidoIx:-MKaQH0d8P4CsZvQjTMn:b-1fxxa3) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/examples/Cassie/osc/walking_speed_control.cc#L68)):*

For code readability, gains should be declared and set in the `.h` files. I think it's fine as is right now, but we once we settle on a foot placement strategy we should remove the other set of gains and move the new ones into the header.
</blockquote></details>

I agree. I also think that gains should be centralized to run_osc_walking_controller.cc and the controller gains file.
___
*[examples/Cassie/osc/walking_speed_control.cc, line 109 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGELvnT1Irc7eXEbXMg:-MKaQwNq0R5N3hstQeMc:b-ph6gcg) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/examples/Cassie/osc/walking_speed_control.cc#L109)):*

nitpick, move the asserts to the start of the function if possible. In this case, there should be a check that the com_port is connected.
</blockquote></details>

It's not always connected depending on the flag `is_using_predicted_com_`, so maybe the current format is fine? 
___
*[systems/controllers/swing_ft_traj_gen.h, line 37 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGENQXn704iP1LO8Ejo:-MKaRTDBARJXdwn-Qg7F:b-4vkp4j) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/systems/controllers/swing_ft_traj_gen.h#L37)):*

Do we need both footstep_offset and center_line_offset? They seem to be accomplishing similar tasks
</blockquote></details>

One is a guard for collision avoidance, and the other is a heuristic offset for control. I found they are both useful although I don't like the heuristic offset. (let me know if you found a better control for the offset!)
___
*[systems/controllers/swing_ft_traj_gen.h, line 61 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGEMZ2zAxyrNRpY-rH5:-MKaSh3A7RTNhFs0JZ6R:b-7rl0jd) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/systems/controllers/swing_ft_traj_gen.h#L61)):*

Add a TODO to clean up these parameters. We should make a decision to either keep or remove these options. 
</blockquote></details>

done.
___
*[systems/controllers/swing_ft_traj_gen.cc, line 300 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGENfno6GoO9tiqbI-q:-MKaT4j1B5KDfWUTxp3N:b-5dup6n) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/systems/controllers/swing_ft_traj_gen.cc#L300)):*

Need to update the generation of the swing foot trajectory to be smooth up to the second derivative. The update should be in the `cassie_experimental` branch.
</blockquote></details>

Good call. Updated.

yangwill (changes requested):
FYI, there are some build errors.

> Reviewed 4 of 5 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/193#-:-MMawgIFAm9WZ_Wnbijr:b-mb1r66)* status: 15 of 16 files reviewed, 1 unresolved discussion (waiting on @yangwill and @yminchen)

___
*[systems/controllers/swing_ft_traj_gen.h, line 37 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGENQXn704iP1LO8Ejo:-MMawDz_3TYshjkB8-zx:bgf3s) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/systems/controllers/swing_ft_traj_gen.h#L37)):*

One is a guard for collision avoidance, and the other is a heuristic offset for control. I found they are both useful although I don't like the heuristic offset. (let me know if you found a better control for the offset!)
</blockquote></details>

I think it is fine as is, but we should think of alternatives. Or maybe combine them both into a single heuristic. Setting them both to 0 seems to actually have the best performance (if we ignore the fact that the feet are on top of each other.

yminchen (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/193)* status: 13 of 16 files reviewed, 1 unresolved discussion (waiting on @yangwill)

___
*[systems/controllers/swing_ft_traj_gen.h, line 37 at r1](https://reviewable.io/reviews/dairlab/dairlib/193#-MGENQXn704iP1LO8Ejo:-MMvlXmd3OoJLkfCMOWp:bskt2xe) ([raw file](https://github.com/dairlab/dairlib/blob/1d1e25702c1583d11240d9091bf9f8ceb124ca1c/systems/controllers/swing_ft_traj_gen.h#L37)):*

I think it is fine as is, but we should think of alternatives. Or maybe combine them both into a single heuristic. Setting them both to 0 seems to actually have the best performance (if we ignore the fact that the feet are on top of each other.
</blockquote></details>

One option is to extract the collision avoidance part into a new leafsystem (it's a TODO in the header file)

yangwill (approved):
LGTM! Will need a new PR after we get the walking controller working on hardware.

> Reviewed 1 of 5 files at r2, 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/193#-:-MMwqSF66jHireR9vcEB:bb536ad)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #194, opened 2020-08-07
Setting standing gains from yaml, setting target height from lcm message

yangwill:
Updating the OSC standing controller from the experimental branch. New features include:
- Setting gains from a yaml file
- Receiving the target center of mass height from an LCM message. Currently sent from an updated pd_panel script in the drake director

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/194)


yangwill:
The walking controller also has gains set by a yaml file but I want to get the walking controller working on hardware before merging the many updates into master.

PR #195, opened 2020-08-12
Adding low pass filter to imu measurement

yangwill:
Adding a low pass filter to the IMU measurement
- interesting note: adding the low pass filter changes the realtime rate from ~0.8 to ~0.6.
- this improves the stability of the tuned walking controller (tested in `cassie_experimental` branch)

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/195)

### DIFF
--- a/examples/Cassie/cassie_utils.cc
+++ b/examples/Cassie/cassie_utils.cc
@@ -7,6 +7,7 @@
 #include "drake/multibody/parsing/parser.h"
 #include "drake/multibody/tree/linear_spring_damper.h"
 #include "drake/multibody/tree/revolute_spring.h"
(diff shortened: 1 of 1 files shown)


yminchen:
The slowdown might come from the fact that the filter is in continuous time (https://github.com/RobotLocomotion/drake/blob/8592c35bd05810323f9b7f46b42542cd53349168/systems/primitives/first_order_low_pass_filter.cc#L27)

I think having a realtime simulation is valuable. Maybe we can implement a discrete version? For example, using `DeclareDiscreteState` and `DeclarePeriodicEvent`

yangwill:
Closing PR because the drake system uses continuous variables and we no longer need this feature.

PR #198, opened 2020-09-25
Fix bugs in dircon_saved_traj and update Drake version

yminchen:
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/198)


yangwill (approved):
Thanks for catching these bugs!

PR #199, opened 2020-10-13
Blend efforts after switch

yangwill:
Blends efforts after a controller switch message. This is to smoothly transition between controllers.

One issue that still needs to be fixed:
- Occasionally the "previous efforts" stored in `discrete_state->get_mutable_vector(prev_efforts_index_)` is written with the new controller efforts instead of the old controller efforts

Open to any suggestions about a fix for this. It looks like there's a race condition, but I don't see a clear way to avoid it.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199)

### DIFF
--- a/examples/Cassie/BUILD.bazel
+++ b/examples/Cassie/BUILD.bazel
@@ -83,6 +83,7 @@ cc_library(
     srcs = ["input_supervisor.cc"],
     hdrs = ["input_supervisor.h"],
     deps = [
+        "//lcmtypes:lcmt_robot",
         "//systems/primitives",
         "@drake//:drake_shared_library",
     ],
--- a/examples/Cassie/dispatcher_robot_in.cc
+++ b/examples/Cassie/dispatcher_robot_in.cc
@@ -72,6 +72,9 @@ int do_main(int argc, char* argv[]) {
                      true /*spring model*/, false /*loop closure*/);
   plant.Finalize();
 
+  // Channel name of the input switch
+  std::string switch_channel = "INPUT_SWITCH";
+
   // Create LCM receiver for commands
   auto command_receiver = builder.AddSystem<RobotInputReceiver>(plant);
 
@@ -79,6 +82,9 @@ int do_main(int argc, char* argv[]) {
   auto state_sub = builder.AddSystem(
       LcmSubscriberSystem::Make<dairlib::lcmt_robot_output>(
           FLAGS_state_channel_name, &lcm_local));
+  auto controller_switch_sub = builder.AddSystem(
+      LcmSubscriberSystem::Make<dairlib::lcmt_controller_switch>(
+          switch_channel, &lcm_local));
   auto state_receiver = builder.AddSystem<systems::RobotOutputReceiver>(plant);
   builder.Connect(*state_sub, *state_receiver);
 
@@ -98,6 +104,8 @@ int do_main(int argc, char* argv[]) {
                   input_supervisor->get_input_port_state());
   builder.Connect(command_receiver->get_output_port(0),
                   input_supervisor->get_input_port_command());
+  builder.Connect(controller_switch_sub->get_output_port(),
+                  input_supervisor->get_input_port_controller_switch());
 
   // Create and connect translator
   auto input_translator =
@@ -127,8 +135,7 @@ int do_main(int argc, char* argv[]) {
   auto owned_diagram = builder.Build();
   owned_diagram->set_name("dispatcher_robot_in");
 
-  // Channel name of the input switch
-  std::string switch_channel = "INPUT_SWITCH";
+
   // Channel names of the controllers
   std::vector<std::string> input_channels;
   input_channels.push_back(FLAGS_control_channel_name_1);
--- a/examples/Cassie/dispatcher_robot_in.cc
+++ b/examples/Cassie/dispatcher_robot_in.cc
@@ -135,7 +135,6 @@ int do_main(int argc, char* argv[]) {
   auto owned_diagram = builder.Build();
   owned_diagram->set_name("dispatcher_robot_in");
 
-
   // Channel names of the controllers
   std::vector<std::string> input_channels;
   input_channels.push_back(FLAGS_control_channel_name_1);
@@ -150,7 +149,7 @@ int do_main(int argc, char* argv[]) {
        command_receiver,
        input_channels,
        FLAGS_control_channel_name_1,
-       switch_channel,
+       controller_switch_sub,
        true);
   loop.Simulate();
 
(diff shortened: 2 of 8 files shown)


mposa:
Does the race condition boil down to wanting to control the sequence between the `output` and `update` function calls? I think we could force `output` to be called after `update` by

1. Re-writing the LCM-driven loop in the more manual format that's now used elsewhere in our code (e.g. https://github.com/DAIRLab/dairlib/blob/master/examples/Cassie/dispatcher_robot_out.cc#L301-L328)
2. Declaring the output function to be "forced"
3. Manually triggering the output.

I'm not sure, however, if this would address your race condition.

yangwill:
I fixed the race condition by using the same `LcmSubscriberSystem` in both the driven loop and the other leaf systems. This is done without having to change the API for LcmDrivenLoop.

yminchen (changes requested):
Per face to face discussion, Will is going to update LcmDrivenLoop a little bit. Additionally, we should add comments about why we add AdvanceTo() to the code and what we haven't figured out so far.

> Reviewed 3 of 6 files at r1, 4 of 4 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199#-:-MKDKv4Z4RQ-MpMkO-AF:b-pxj6mj)* status: all files reviewed, 1 unresolved discussion (waiting on @yangwill)

___
*[examples/Cassie/run_controller_switch.cc, line 17 at r2](https://reviewable.io/reviews/dairlab/dairlib/199#-MKBKovH3xI9jMFBWdjc:-MKBKovH3xI9jMFBWdjd:bmeitvj) ([raw file](https://github.com/dairlab/dairlib/blob/f2d171266f26c410e77367e88f3d6bb2108316ca/examples/Cassie/run_controller_switch.cc#L17)):*
> ```cpp
> using drake::systems::lcm::TriggerTypeSet;
> 
> DEFINE_string(channel_x, "CASSIE_STATE_DISPATCHER",
> ```

Since we are changing the default channel name, can we make sure that the procman scripts in `cassie_simulation.pmd` still work?

yangwill (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @yminchen)

___
*[examples/Cassie/run_controller_switch.cc, line 17 at r2](https://reviewable.io/reviews/dairlab/dairlib/199#-MKBKovH3xI9jMFBWdjc:-MO7UDmo5aGPn-VABZaV:bmoe433) ([raw file](https://github.com/dairlab/dairlib/blob/f2d171266f26c410e77367e88f3d6bb2108316ca/examples/Cassie/run_controller_switch.cc#L17)):*

Since we are changing the default channel name, can we make sure that the procman scripts in `cassie_simulation.pmd` still work? 
</blockquote></details>

I just checked the procman script, and no changes need to be made.

yangwill (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199)* status: 0 of 8 files reviewed, 1 unresolved discussion (waiting on @yminchen)

___
*[examples/Cassie/run_controller_switch.cc, line 17 at r2](https://reviewable.io/reviews/dairlab/dairlib/199#-MKBKovH3xI9jMFBWdjc:-MOCOHWaBgof9fezClk0:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/f2d171266f26c410e77367e88f3d6bb2108316ca/examples/Cassie/run_controller_switch.cc#L17)):*

I just checked the procman script, and no changes need to be made.
</blockquote></details>

Done.

yminchen (changes requested):
Just two minor comments

> Reviewed 6 of 8 files at r3.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199#-:-MODNFD63obYGA69Kxt1:b-3sru0u)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @yangwill and @yminchen)

___
*[systems/framework/lcm_driven_loop.h, line 148 at r2](https://reviewable.io/reviews/dairlab/dairlib/199#-MODDtvh7tR2KDjJbiix:-MODDtvh7tR2KDjJbiiy:b-dw44xn) ([raw file](https://github.com/dairlab/dairlib/blob/f2d171266f26c410e77367e88f3d6bb2108316ca/systems/framework/lcm_driven_loop.h#L148)):*
> ```objc
>                       "", is_forced_publish){};
> 
>   // Getters for diagram and its context
> ```

Can we retain these getters? Although we might not be using them in master, I'm using them for some application in my branch. 
___
*[systems/framework/lcm_driven_loop.h, line 253 at r3](https://reviewable.io/reviews/dairlab/dairlib/199#-MODNP0P5uhAkDRvogxN:-MODNP0P5uhAkDRvogxO:brh7im2) ([raw file](https://github.com/dairlab/dairlib/blob/b4a84b1343a76ce2d4fc40332761636733929cd9/systems/framework/lcm_driven_loop.h#L253)):*
> ```objc
>         // received in the leaf systems without the encountering a race
>         // condition. Still need to investigate the exact sequence that causes
>         // the race condition
> ```

Can you add a little more background explaining that we need this because we have two subscriber listening to the same lcm channel? And although it's possible to use only subscriber, we would have to compromise the readability.

yangwill (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199)* status: 6 of 8 files reviewed, 2 unresolved discussions (waiting on @yminchen)

___
*[systems/framework/lcm_driven_loop.h, line 148 at r2](https://reviewable.io/reviews/dairlab/dairlib/199#-MODDtvh7tR2KDjJbiix:-MODP8r_-tmb7b1ZjlwB:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/f2d171266f26c410e77367e88f3d6bb2108316ca/systems/framework/lcm_driven_loop.h#L148)):*

Can we retain these getters? Although we might not be using them in master, I'm using them for some application in my branch. 
</blockquote></details>

Done.
___
*[systems/framework/lcm_driven_loop.h, line 253 at r3](https://reviewable.io/reviews/dairlab/dairlib/199#-MODNP0P5uhAkDRvogxN:-MODPA5PEpmsGQBqfh9r:bb6un1j) ([raw file](https://github.com/dairlab/dairlib/blob/b4a84b1343a76ce2d4fc40332761636733929cd9/systems/framework/lcm_driven_loop.h#L253)):*

Can you add a little more background explaining that we need this because we have two subscriber listening to the same lcm channel? And although it's possible to use only subscriber, we would have to compromise the readability.  
</blockquote></details>

Done

yminchen (approved):
:lgtm:

> Reviewed 1 of 8 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/199#-:-MODQ44I4RjEBdVNixsj:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #201, opened 2020-10-13
adding edge case for exponentialpluspiecewisepolynomial

yangwill:
`ExponentialPlusPiecewisePolynomial` extends from `PiecewiseTrajectory` and not `PiecewisePolynomial` and therefore does not support `EvalDerivative`. This adds an edge case just for `ExponentialPlusPiecewisePolynomial` which is used by `lipm_traj_gen.cc`

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/201)

### DIFF
--- a/systems/controllers/osc/osc_tracking_data.cc
+++ b/systems/controllers/osc/osc_tracking_data.cc
@@ -55,8 +55,14 @@ bool OscTrackingData::Update(
     // Careful: must update y_des_ before calling UpdateYAndError()
     // Update desired output
     y_des_ = traj.value(t);
-    ydot_des_ = traj.EvalDerivative(t, 1);
-    yddot_des_ = traj.EvalDerivative(t, 2);
+    if(traj.has_derivative()){
+      ydot_des_ = traj.EvalDerivative(t, 1);
+      yddot_des_ = traj.EvalDerivative(t, 2);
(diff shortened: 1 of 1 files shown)


mposa:
Seems OK--though if this is a computational bottleneck we could easily add `EvalDerivative` to `ExponentialPlusPiecewisePolynomial`

yangwill:
I'm not sure we could avoid the main computational bottleneck which is the derivative of the exponential part (https://github.com/RobotLocomotion/drake/blob/master/common/trajectories/exponential_plus_piecewise_polynomial.cc#L44), the piecewise polynomial part is quick for our code because I think we only construct one segment.

I agree that this is a ugly fix though.

yminchen (approved):
I approve this PR so that the walking controller in master branch doesn't throw a segfault error, but we should add a TODO mentioning that the code can be cleaned up if we add `EvalDerivative` to `ExponentialPlusPiecewisePolynomial`

> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/201#-:-MKB8juA95ZJ6bkL6Xz3:bhetrs7)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #203, opened 2020-10-23
Fix a bug for fixed base cassie state estimator

yminchen:
There was a bug in `cassie_state_estimator` when we run the fixed based `dispatcher_robot_out`

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/203)


yangwill (approved):
LGTM

PR #212, opened 2020-12-06
Fix a bug in lipm_traj_gen

yminchen:
A little bug caught by `DRAKE_ASSERT()` when I run the code with valgirnd.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/212)

### DIFF
--- a/systems/controllers/lipm_traj_gen.cc
+++ b/systems/controllers/lipm_traj_gen.cc
@@ -63,7 +63,7 @@ LIPMTrajGenerator::LIPMTrajGenerator(
   PiecewisePolynomial<double> pp_part(VectorXd(0));
   MatrixXd K = MatrixXd::Ones(0, 0);
   MatrixXd A = MatrixXd::Identity(0, 0);
-  MatrixXd alpha = MatrixXd::Ones(0, 0);
+  MatrixXd alpha = MatrixXd::Ones(0, 1);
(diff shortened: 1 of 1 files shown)


yangwill:
Can you give an explanation for what the bug is? Why don't the initializations for K and A not cause the same error?

yminchen:
The number of polynomial segments should be equal to the number of columns of `alpha`. 
https://github.com/RobotLocomotion/drake/blob/master/common/trajectories/exponential_plus_piecewise_polynomial.h#L45
When running in debug mode, we get a runtime error from this line of code in the link.

This bug might cause issues, but it's still better to remove it to prevent potential memory leak.

yangwill:
I'm just asking because I'm not convinced that this should be the source of the bug.

yminchen:
Hmm maybe my statement was confusing. I probably shouldn't call this a "bug". 
I was debugging my code in a branch, but the program stopped early because of the `DRAKE_ASSERT()`. I had to make this one line fix to see valgrind outputing the message of the actual bug.

yangwill:
Another option is to remove `K` `A` and `alpha` and just use the default constructor - ie `ExponentialPlusPiecewisePolynomial<double> exp;`. This fix seems cleaner to me than setting a value for alpha and not using it later.

yminchen:
Yeah using default constructor is much cleaner!

yangwill (approved):
LGTM!

PR #227, opened 2021-01-31
Merge cassie experimental

yangwill:
This PR updates the walking and standing controller with adjustments from the hardware experiments.

Cassie is able to reliably transition from PD setpoints to a balancing position and finally to the walking controller.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/227)


yangwill:
@yminchen, I was able to fix the end time issue when constructing the LIPM trajectory, but the fix is not entirely simple. Part of the cause is the `context` time being behind the robot's time and the other is because of the randomness of when each leaf system calls its discrete variable updates.

A simple solution is to always add a few milliseconds to the end times, this is slightly cleaner but still hacky.

yminchen (approved):
Looks good to me! (I'm assuming there is no change since our merge on Friday) 

> Reviewed 21 of 21 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/227#-:-MSP0wcM-WHCrOc4GmjB:b-ljru9a)* status: :shipit: complete! all files reviewed, all discussions resolved

yangwill:
We still need a decision on how we want to deal with the start/end time when constructing trajectories. I propose enforcing the start end is at least 1ms before the end time (this way the target trajectory at the end time is guaranteed to be the desired value).

PR #231, opened 2021-02-07
Properly handle discontinuities when adding velocity cost

ShaneRozenLevy:
`doAddRunningCost` will not properly handle the discontinuities in velocity from impacts. I added a function which will only add a cost on sum of velocity squared, but will properly handle discontinuities. Code was tested on optimizing leaps for Spirit.

### DIFF
--- a/systems/trajectory_optimization/dircon/dircon.cc
+++ b/systems/trajectory_optimization/dircon/dircon.cc
@@ -576,6 +576,36 @@ void Dircon<T>::DoAddRunningCost(const drake::symbolic::Expression& g) {
           h_vars()(N() - 2) / 2);
 }
 
+template<typename T>
+void Dircon<T>::AddVelocityCost(const double velocityCostGain){
+  double n_v = plant_.num_velocities();
+
+  // Add velocity cost handling discontinuities
+  // Loop through each mode and each knot point and use trapezoidal integration
+  for(int mode_index = 0; mode_index < num_modes(); mode_index++){
+    for(int knot_index = 0; knot_index < mode_length(mode_index)-1; knot_index++){
+
+      // Get lower and upper knot velocities
+      drake::solvers::VectorXDecisionVariable vel  = state_vars(mode_index, knot_index).tail(n_v);
+      drake::solvers::VectorXDecisionVariable vel_up = state_vars(mode_index, knot_index+1).tail(n_v);
+
+      drake::symbolic::Expression hi = timestep(mode_start_[mode_index] + knot_index)[0];
+
+      // Loop through and calculate sum of velocities squared
+      drake::symbolic::Expression vel_sq = 0;
+      drake::symbolic::Expression vel_sq_up = 0;
+      for(int i = 0; i< n_v; i++){
+        vel_sq += vel[i] * vel[i];
+        vel_sq_up += vel_up[i] * vel_up[i];
+      }
+
+      // Add cost
+      AddCost(hi/2.0 * velocityCostGain * (vel_sq + vel_sq_up));
+    }
+  }
+
+}
+
 template <typename T>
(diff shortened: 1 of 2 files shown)


yangwill (changes requested):
> Reviewed 2 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/231)* status: all files reviewed, 3 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[systems/trajectory_optimization/dircon/dircon.h, line 191 at r1](https://reviewable.io/reviews/dairlab/dairlib/231#-MVassXQDBJgTsQSaYje:-MVassXRCZTAsLi1Xw1Z:bcgvpx4) ([raw file](https://github.com/dairlab/dairlib/blob/ee8f74a8c8a2e2791f75fc512d514807cf00ffa3/systems/trajectory_optimization/dircon/dircon.h#L191)):*
> ```objc
> 
>   /// Adds cost on sum of velocities squared while properly handling the
>   /// discontinuities from impacts
> ```

Write out the cost expression if possible, see drake documentation: https://drake.mit.edu/doxygen_cxx/classdrake_1_1solvers_1_1_mathematical_program.html#ab5d909ab79bbc45767206daaa8d145c8 for examples.
___
*[systems/trajectory_optimization/dircon/dircon.h, line 192 at r1](https://reviewable.io/reviews/dairlab/dairlib/231#-MVardx64Lf0zGECazJA:-MVardx64Lf0zGECazJB:b-pcvnbv) ([raw file](https://github.com/dairlab/dairlib/blob/ee8f74a8c8a2e2791f75fc512d514807cf00ffa3/systems/trajectory_optimization/dircon/dircon.h#L192)):*
> ```objc
>   /// Adds cost on sum of velocities squared while properly handling the
>   /// discontinuities from impacts
>   void AddVelocityCost(const double velocityCostGain);
> ```

codestyle: parameter should be velocity_cost_gain or perhaps a matrix so the cost would be J = v^T Q v.
___
*[systems/trajectory_optimization/dircon/dircon.cc, line 600 at r1](https://reviewable.io/reviews/dairlab/dairlib/231#-MVauV09A04wj8AtzCu8:-MVauV09A04wj8AtzCu9:b-3x3xyq) ([raw file](https://github.com/dairlab/dairlib/blob/ee8f74a8c8a2e2791f75fc512d514807cf00ffa3/systems/trajectory_optimization/dircon/dircon.cc#L600)):*
> ```cpp
>         vel_sq += vel[i] * vel[i];
>         vel_sq_up += vel_up[i] * vel_up[i];
>       }
> ```

How much more effort would it be to allow any expression of the decision variables rather than just the squared velocity cost? I can take a stab at it.

ShaneRozenLevy (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/231)* status: all files reviewed, 3 unresolved discussions (waiting on @ShaneRozenLevy and @yangwill)

___
*[systems/trajectory_optimization/dircon/dircon.cc, line 600 at r1](https://reviewable.io/reviews/dairlab/dairlib/231#-MVauV09A04wj8AtzCu8:-MVazZ-r2Psn3jr3O4CL:b-u4bclw) ([raw file](https://github.com/dairlab/dairlib/blob/ee8f74a8c8a2e2791f75fc512d514807cf00ffa3/systems/trajectory_optimization/dircon/dircon.cc#L600)):*

How much more effort would it be to allow any expression of the decision variables rather than just the squared velocity cost? I can take a stab at it.
</blockquote></details>

You should take a stab at it. I don't know how I would do that since it would need to know when to replace the velocity terms with the preimpact velocities and when to replace the velocities with the post impact velocities. 

There already is a function which can take an arbitrary cost function, but it always replaces the velocities with the preimpact velocities.

PR #233, opened 2021-03-16
Fix a bug of using stale context time in FiniteStateMachineEventTime

yminchen:
It's the same issue described in #63, but I didn't fix it with the same approach as #188, because we would have to set the message time manually which makes LcmDrivenLoop more complicated.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233)

### DIFF
--- a/examples/Cassie/run_osc_walking_controller.cc
+++ b/examples/Cassie/run_osc_walking_controller.cc
@@ -207,13 +207,13 @@ int DoMain(int argc, char* argv[]) {
                                             right_stance_state};
   auto liftoff_event_time =
       builder.AddSystem<systems::FiniteStateMachineEventTime>(
-          single_support_states);
+          plant_w_spr, single_support_states);
   liftoff_event_time->set_name("liftoff_time");
   builder.Connect(fsm->get_output_port(0),
                   liftoff_event_time->get_input_port_fsm());
   auto touchdown_event_time =
       builder.AddSystem<systems::FiniteStateMachineEventTime>(
-          std::vector<int>(1, double_support_state));
+          plant_w_spr, std::vector<int>(1, double_support_state));
   touchdown_event_time->set_name("touchdown_time");
   builder.Connect(fsm->get_output_port(0),
                   touchdown_event_time->get_input_port_fsm());
--- a/examples/Cassie/run_osc_walking_controller.cc
+++ b/examples/Cassie/run_osc_walking_controller.cc
@@ -211,12 +211,16 @@ int DoMain(int argc, char* argv[]) {
   liftoff_event_time->set_name("liftoff_time");
   builder.Connect(fsm->get_output_port(0),
                   liftoff_event_time->get_input_port_fsm());
+  builder.Connect(simulator_drift->get_output_port(0),
+                  liftoff_event_time->get_input_port_state());
   auto touchdown_event_time =
       builder.AddSystem<systems::FiniteStateMachineEventTime>(
           plant_w_spr, std::vector<int>(1, double_support_state));
   touchdown_event_time->set_name("touchdown_time");
   builder.Connect(fsm->get_output_port(0),
                   touchdown_event_time->get_input_port_fsm());
+  builder.Connect(simulator_drift->get_output_port(0),
+                  touchdown_event_time->get_input_port_state());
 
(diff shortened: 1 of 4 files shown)


yminchen:
On a second thought, maybe LcmDrivenLoop is not going to be much more complicated if I set the message time manually. I will make a new PR later.

yminchen:
Nvm, we need a public function `set_next_message_time` which makes the generalization of LcmDrivenLoop not trivial.

yminchen:
I realized that we can just get the message time directly from robot output. The case here is different form #188 where the lcm parser doesn't populate the message time.

yminchen:
@yangwill could you review this if you are available?

yangwill (changes requested):
Looks fine, but would probably move the state input to the `time_based_fsm` and then have the FSM have an output port with the current time (and have the fsm_event_time use that output. It's not obvious why the event time listener should need to know state information. It is reasonable for the FSM to take in the robot state.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233#-:-MWHv6F_-eJ9paR6EwPT:b-t05cb3)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mposa and @yminchen)

___
*[systems/controllers/fsm_event_time.cc, line 89 at r3](https://reviewable.io/reviews/dairlab/dairlib/233#-MWHuumaApZwoOTkj5X1:-MWHuumbDeLsg3sxvvqX:b4kzskg) ([raw file](https://github.com/dairlab/dairlib/blob/8903acd2bdcd0c860bae4853066444620a9e47b7/systems/controllers/fsm_event_time.cc#L89)):*
> ```cpp
>       discrete_state->get_mutable_vector(prev_time_of_state_of_interest_idx_)
>               .get_mutable_value()
>           << static_cast<double>(robot_output->get_timestamp());
> ```

Is the static_cast necessary?

yangwill (commented):
Otherwise LGTM!

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233#-:-MWHvmcb1BXYfaC3sIO1:b-d665z8)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mposa and @yminchen)

yminchen (commented):
Yeah it's easy to think that we are using the robot state when using `RobotOutputVector`, although it contains more than just that. 
Creating a `current_time` output in `time_based_fsm` doesn't look straightforward to me. If we were to create a current_time output, maybe it's best to create a demux-looking block that takes in robot output and spits out time?

I could also improve the documentation of the event time block. 

What do you think? 

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233#-:-MWJb8D8DfFNRpb6o_cw:b8ve42d)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mposa and @yminchen)

yminchen (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233)* status: 0 of 4 files reviewed, 1 unresolved discussion (waiting on @mposa and @yangwill)

___
*[systems/controllers/fsm_event_time.cc, line 89 at r3](https://reviewable.io/reviews/dairlab/dairlib/233#-MWHuumaApZwoOTkj5X1:-MWKweLi7qqMyMphKMzE:b6xey1r) ([raw file](https://github.com/dairlab/dairlib/blob/8903acd2bdcd0c860bae4853066444620a9e47b7/systems/controllers/fsm_event_time.cc#L89)):*

Is the static_cast necessary? 
</blockquote></details>

Nope, I don't think it's necessary here. I just removed the casting.

yangwill (approved):
> Reviewed 3 of 4 files at r2, 1 of 1 files at r3, 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mposa)

yangwill (approved):
:lgtm:

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/233#-:-MWL6rPW64-VLTY5X1W-:bnfp4nl)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @mposa)

PR #234, opened 2021-03-21
Manual estop

yangwill:
Implementing the soft-estop (which applies a damping only controller) in the input supervisor.

The soft-estop is triggered by the radio and the dispatcher will remain in the error state after it is triggered.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/234)

### DIFF
--- a/examples/Cassie/input_supervisor.cc
+++ b/examples/Cassie/input_supervisor.cc
@@ -76,6 +76,7 @@ InputSupervisor::InputSupervisor(
   K_(7, 13) = -1;
   K_(8, 19) = -1;
   K_(9, 21) = -1;
+  K_ = 10 * K_;
 
   // Create update for error flag
   DeclarePeriodicDiscreteUpdateEvent(update_period, 0,
--- a/examples/Cassie/input_supervisor.cc
+++ b/examples/Cassie/input_supervisor.cc
@@ -1,5 +1,7 @@
 #include "examples/Cassie/input_supervisor.h"
 
+#include <dairlib/lcmt_cassie_out.hpp>
+
 #include "dairlib/lcmt_controller_switch.hpp"
 #include "systems/framework/output_vector.h"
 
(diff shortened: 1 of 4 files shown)


examples/Cassie/input_supervisor.cc
```diff
@@ -58,6 +65,11 @@ InputSupervisor::InputSupervisor(
   switch_time_index_ = DeclareDiscreteState(1);
   prev_efforts_time_index_ = DeclareDiscreteState(1);
   prev_efforts_index_ = DeclareDiscreteState(num_actuators_);
+  soft_estop_flag_index_ = DeclareDiscreteState(1);
+
+  K_ = plant_.MakeActuationMatrix().transpose();
+  K_ *= kEStopGain;
+  std::cout << K_ << std::endl;
```
Brian-Acosta, line 72: If we're going to leave this in let's also print out "Soft E-stop Gains:" or similar

examples/Cassie/input_supervisor.cc
```diff
@@ -77,13 +94,23 @@ void InputSupervisor::SetMotorTorques(const Context<double>& context,
       is_error || (command->get_timestamp() -
                        context.get_discrete_state(prev_efforts_time_index_)[0] >
                    kMaxControllerDelay);
+  is_error =
+      is_error || context.get_discrete_state(soft_estop_flag_index_)[0] != 0;
```
Brian-Acosta, line 98: By adding the soft e-stop as an error we are setting the output command to 0 instead of the damping command aren't we?

Brian-Acosta (commented):
Looks good for the most part - I'm not sure that we want to include soft e-stop as an error for the reasons mentioned in [this comment](https://github.com/DAIRLab/dairlib/pull/234/files#r598345810).

examples/Cassie/input_supervisor.cc
yangwill: oops meant to remove this

examples/Cassie/input_supervisor.cc
yangwill: this is not obvious from the code and probably could be more obvious but since the soft e-stop check and override also has a return statement, the lines that zero out the controller commands don't get called.

Brian-Acosta: Ah I see it now. I couldn't tell because of how the intervening code was hidden.

Brian-Acosta (commented):
LGTM, can merge after testing on hardware this week

Brian-Acosta (approved):
Works!

PR #235, opened 2021-03-22
Handling velocitiy discontinuities correctly in Dircon cost

yangwill:
Implementation of a more general version of #231.

Creates a `Substitution` member variable in Dircon that replaces the velocity variables with the correct post-impact velocity variables.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/235)


yangwill:
Oops this was branched off from the wrong branch.

PR #236, opened 2021-03-22
addressing velocity discontinuities correctly in the cost function

yangwill:
Implementation of a more general version of #231.

Creates a Substitution member variable in Dircon that replaces the velocity variables with the correct post-impact velocity variables.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/236)

### DIFF
--- a/systems/trajectory_optimization/dircon/dircon.cc
+++ b/systems/trajectory_optimization/dircon/dircon.cc
@@ -11,6 +11,7 @@ namespace trajectory_optimization {
 using drake::VectorX;
 using drake::multibody::MultibodyPlant;
 using drake::solvers::MathematicalProgramResult;
+using drake::solvers::MatrixXDecisionVariable;
 using drake::solvers::VectorXDecisionVariable;
 using drake::symbolic::Expression;
 using drake::systems::Context;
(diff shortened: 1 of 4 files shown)


systems/trajectory_optimization/dircon/dircon.cc
```diff
@@ -102,6 +103,11 @@ Dircon<T>::Dircon(std::unique_ptr<DirconModeSequence<T>> my_sequence,
     if (i_mode > 0) {
       v_post_impact_vars_.push_back(NewContinuousVariables(
           plant_.num_velocities(), "v_p[" + std::to_string(i_mode) + "]"));
+      v_post_impact_vars_substitute_.push_back(
+          {state().tail(plant_.num_velocities()),
+           Eigen::Map<MatrixXDecisionVariable>(
+               v_post_impact_vars_.back().data(), plant_.num_velocities(), 1)
+               .cast<Expression>()});
```
mposa, line 110: Why use Eigen::Map here? An alternate option would be 
```
v_post_impact_vars_.back().tail(plant_.num_velocities()).template cast<Expression>()}
```

It's a bit easier to read, though without the memory efficiency of Map.

systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -107,6 +109,17 @@ HybridDircon<T>::HybridDircon(const MultibodyPlant<T>& plant,
           constraints_[i]->countConstraintsWithoutSkipping(),
           "impulse[" + std::to_string(i) + "]"));
     }
+    //  Post-impact variables. Note: impulse variables declared below.
+    if (i > 0) {
+      v_post_impact_vars_substitute_.push_back(
```
mposa, line 114: See comment on dircon.cc

mposa (approved):
Looks good to me. Just the one comment/suggestion.

systems/trajectory_optimization/dircon/dircon.cc
yangwill: I was using https://github.com/RobotLocomotion/drake/blob/a3312d0f2d526c074f793c0c7eba7f41153c7055/systems/trajectory_optimization/multiple_shooting.cc#L95 as a reference. Though I'm not confident if `template_cast` wouldn't achieve the same effect.

PR #237, opened 2021-03-24
Set impact velocities initial guess

ShaneRozenLevy:
Added new function for setting initial guess from trajectories. This has a few advantages over the current interface:

1. Sets initial guess for post impact velocities
2. Sets initial guess for the duration of each segment based the length of time allocated per mode 
3. When evaluating force trajectory & state trajectory uses the correct times

I have done some testing of this and it seems good, but it likely needs more testing and a second set of eyes before merging it in.

### DIFF
--- a/systems/trajectory_optimization/dircon/dircon.cc
+++ b/systems/trajectory_optimization/dircon/dircon.cc
@@ -643,12 +643,12 @@ template <typename T>
 void Dircon<T>::SetInitialForceTrajectory(
     int mode_index, const PiecewisePolynomial<double>& traj_init_l,
     const PiecewisePolynomial<double>& traj_init_lc,
-    const PiecewisePolynomial<double>& traj_init_vc) {
+    const PiecewisePolynomial<double>& traj_init_vc,
+    const double start_time) {
   const auto& mode = get_mode(mode_index);
-  double start_time = 0;
   double h;
   if (timesteps_are_decision_variables())
-    h = GetInitialGuess(h_vars()[0]);
+    h = GetInitialGuess(h_vars()[get_mode_start(mode_index)]);
   else
     h = fixed_timestep();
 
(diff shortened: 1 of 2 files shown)


yangwill (commented):
> Reviewed 1 of 2 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/237)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[systems/trajectory_optimization/dircon/dircon.h, line 139 at r1](https://reviewable.io/reviews/dairlab/dairlib/237#-MWil_j94e_CR34dzxU6:-MWil_jAEznrx8E2SB4O:buyfj2j) ([raw file](https://github.com/dairlab/dairlib/blob/ab86ec73f37b2a461091b4e30921dfe1f08efa30/systems/trajectory_optimization/dircon/dircon.h#L139)):*
> ```objc
>   /// @param start_time the start time for the mode
>   /// @param end_time the end time for a mode
>   void SetInitialDiscontinuousTrajectory(
> ```

Can we rename this to `SetInitialTrajectoryForMode()`?
___
*[systems/trajectory_optimization/dircon/dircon.h, line 142 at r1](https://reviewable.io/reviews/dairlab/dairlib/237#-MWim3HlBeRkZv2dVDVX:-MWim3HlBeRkZv2dVDVY:b-pvaic1) ([raw file](https://github.com/dairlab/dairlib/blob/ab86ec73f37b2a461091b4e30921dfe1f08efa30/systems/trajectory_optimization/dircon/dircon.h#L142)):*
> ```objc
>       int mode_index, const drake::trajectories::PiecewisePolynomial<double>& traj_init_x,
>       const drake::trajectories::PiecewisePolynomial<double>& traj_init_u,
>       const double start_time,
> ```

Are `start_time` and `end_time` for this trajectory optimization problem or for the supplied state and input trajectory? I feel like there should be a way to automatically retrieve these values
___
*[systems/trajectory_optimization/dircon/dircon.cc, line 738 at r1](https://reviewable.io/reviews/dairlab/dairlib/237#-MWimwn10MVGgp70sSLO:-MWimwn10MVGgp70sSLP:bfubxa9) ([raw file](https://github.com/dairlab/dairlib/blob/ab86ec73f37b2a461091b4e30921dfe1f08efa30/systems/trajectory_optimization/dircon/dircon.cc#L738)):*
> ```cpp
>     // State
>     VectorXd guess_state(state_vars(mode_index,i).size());
>     guess_state = traj_init_x.value(start_time + i * h);
> ```

you can do this in one line, `VectorXd guess_state = traj_init_x.value(state_time + i * h);` 

Same with `guess_control`. 
___
*[systems/trajectory_optimization/dircon/dircon.cc, line 743 at r1](https://reviewable.io/reviews/dairlab/dairlib/237#-MWinE4_CnqFMAT0oPL5:-MWinE4_CnqFMAT0oPL6:b8hvexe) ([raw file](https://github.com/dairlab/dairlib/blob/ab86ec73f37b2a461091b4e30921dfe1f08efa30/systems/trajectory_optimization/dircon/dircon.cc#L743)):*
> ```cpp
>     // Control
>     VectorXd guess_control(input_vars(mode_index, i).size());
>     guess_control = traj_init_u.value(start_time + i * h);
> ```

Would also suggest renaming `guess_state` and `guess_control` to `x_guess` and `u_guess`.

yangwill (commented):
Looks great, just a few minor comments. Thanks for adding this feature!

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/237#-:-MWio5q63W6k96YVgPyn:b-cgc3uo)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @ShaneRozenLevy)

ShaneRozenLevy:
In terms of the start and end time, I wrote this function assuming the start and end of the spline might not be the start and end of mode. If you think thats a reasonable assumption I can set start time and end time using the inputted spline.

yangwill (commented):
I thought it might be possible to get the start and end time of the mode from the member variables. But now I see why it's not necessarily the case if the timesteps are left as decision variables. I would add in the description for `start_time` and `end_time` that these are guesses for the start and end time of the mode.

Please correct me if I'm misunderstanding the purpose of these variables.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/237#-:-MWz6K4JDyqKsaVtRf6m:b-ebp9zn)* status: 1 of 2 files reviewed, 4 unresolved discussions (waiting on @ShaneRozenLevy)

ShaneRozenLevy:
Hmm, so thats close to the purpose of those variables, but not quite it. I use the start and end time for 2 purposes. 

1. To calculate the initial guess for the duration of the mode
2. When evaluating traj_init_x for the intial guess for a particular mode I evaluate it starting at start_time and going to end_time.

Depending on how your code is setup you might want to set start time to be 0 for each mode or maybe you want to set start_time to be your actual guess for the time the mode start. It all depends on which interval of time from the traj_init_x you want to use for the initial guess. I'm also assuming the traj_init_x does not necessarily start and start_time and end at end_time (if I did, I could get start_time and end_time from the pp_traj object).

PR #239, opened 2021-04-15
adding a check to make sure we are not sending NaN in our motor commands

yangwill:
On April 14, we noticed that it is possible to forward NaN motor commands through the input supervisor, resulting in undesirable behavior on Cassie. This change triggers the soft-estop if any value in the motor commands is NaN.

We shouldn't be sending NaNs from our controllers anyway, but this acts as another safety check.

I also made minor refactoring to input supervisor to improve readability, open to feedback.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/239)


Brian-Acosta (commented):
If I'm reading this correctly, is_nan is the only error condition that doesn't also set the is_error flag. Is this the behavior we want?

yangwill:
Updated it so that sending a NaN motor command triggers an error and sets the motor commands to 0.

Brian-Acosta (approved):
LGTM

PR #240, opened 2021-04-21
Fixing indexing bug in hybrid dircon

yangwill:
This PR fixes an off by one indexing error from #236.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/240)

### DIFF
--- a/systems/trajectory_optimization/hybrid_dircon.cc
+++ b/systems/trajectory_optimization/hybrid_dircon.cc
@@ -108,14 +108,12 @@ HybridDircon<T>::HybridDircon(const MultibodyPlant<T>& plant,
       impulse_vars_.push_back(NewContinuousVariables(
           constraints_[i]->countConstraintsWithoutSkipping(),
           "impulse[" + std::to_string(i) + "]"));
-    }
-    //  Post-impact variables. Note: impulse variables declared below.
-    if (i > 0) {
+      //  Post-impact variables. Note: impulse variables declared below.
       v_post_impact_vars_substitute_.push_back(
           {state().tail(plant_.num_velocities()),
            Eigen::Map<MatrixXDecisionVariable>(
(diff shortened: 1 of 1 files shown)


systems/trajectory_optimization/hybrid_dircon.cc
```diff
@@ -108,14 +108,12 @@ HybridDircon<T>::HybridDircon(const MultibodyPlant<T>& plant,
       impulse_vars_.push_back(NewContinuousVariables(
           constraints_[i]->countConstraintsWithoutSkipping(),
           "impulse[" + std::to_string(i) + "]"));
-    }
-    //  Post-impact variables. Note: impulse variables declared below.
-    if (i > 0) {
```
mposa, line 113: I'm not seeing the off-by-one bug here. Isn't some sort of `if` statement necessary here, since there is one fewer impact than the number of modes.?

The old code seems like it's going to build `v_post_impact_vars_substitute_` with all of the `v_post_impact_vars_by_mode` elements, from 0,...,n-2. Won't the new code try to add post impact variables for the last mode (which doesn't exist)?

yangwill (commented):
The fix is the `i-1` change. `v_post_impact_vars_by_mode(0)` has the post-impact velocities variables for the start of mode 1. Which is what we want.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/240#-:-MYrJqCz1yyG5OBobo_i:b-3erb3l)* status: 0 of 1 files reviewed, all discussions resolved

mposa (approved):
Whoops. Misread the modification to the `if` 

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/240#-:-MYrNA-92a-6fQKzlg7t:bon5gn4)* status: 0 of 1 files reviewed, all discussions resolved

PR #247, opened 2021-07-14
Fix spring deflection calculation

yminchen:
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/247)


Brian-Acosta (approved):
LGTM

PR #250, opened 2021-07-23
Update to drake master

mposa:
-Fix deprecation warnings for missing port names. **I chose names that seemed appropriate, but please weigh in if you have thoughts.**
-Remove dead serializer (UDPout) code
-Add Clone method to serializer (UDPin)
-Due to upstream changes, the new method to compute orthonormal bases from a single vector is now `RotationMatrix::MakeFromOneVector`. This does not necessarily produce the same results as the old function (since there is ambiguity in how the other two axes are defined). We used this in `WorldPointEvaluator`. I've updated the unit test, which has manually defined values, but it's worth being aware that this might have unintended consequences if we make assumptions about how those other two axes are ultimately specified.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/250)


mposa:
This is ready for review, if anyone can take a look.  Notable changes are in the PR description.

yangwill (changes requested):
Couple thoughts on ports naming rules.

- Because InputPorts and OutputPorts are treated separately, we don't need `_in` or `_out` suffixes. Even if they are of the same type
- Ports that are lcmtypes must be named as their lcmtype unless there are multiple input/output ports of the same type. If that's the case, add a short suffix that distinguishes the different ports
- Output ports that are trajectories should describe the body that the trajectory is describing and the value. Several examples:
   -   Vertical position of the pelvis should be `pelvis_z`
   -   Hip yaw angle should be `hip_yaw_ang`
   -   Cartesian position of the center of mass should be `com_xyz`
   -   Pelvis orientation in SO(3) should be `pelvis_rot` (not entirely sure about this one)
- Vector-valued ports that are present in the robot dynamics equations should use the same letters (eg. `u`, `x`, `t`, `q`)

I'm not sure about how to name ports that are of type `OutputVector`, `robot_output` feels vague but too wordy at the same time. We should probably distinguish between ports of type `BasicVector`, `TimestampedVector`, and `OutputVector` in the names, so that when we draw the diagram we can easily see which values are being passed along.

Once we agree on a naming convention, I can go through and update all the port names

> Reviewed 46 of 46 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/250#-:-MfYRv7G7IKejuwfRcZa:b-raqsjg)* status: all files reviewed, 4 unresolved discussions (waiting on @mposa)

___
*[examples/Cassie/input_supervisor.cc, line 39 at r1](https://reviewable.io/reviews/dairlab/dairlib/250#-MfYPOfS6fmolEnsfZka:-MfYPOfT5HVCKFbHP5EV:b9yrj52) ([raw file](https://github.com/dairlab/dairlib/blob/fd188fcb7e87e52506a96dff7e1be66a072393c3/examples/Cassie/input_supervisor.cc#L39)):*
> ```cpp
>   state_input_port_ =
>       this->DeclareVectorInputPort(
>               "robbot_output",
> ```

typo in robbot
___
*[examples/Cassie/sim_cassie_sensor_aggregator.cc, line 37 at r1](https://reviewable.io/reviews/dairlab/dairlib/250#-MfYPg0t4sj-FMLwRv1g:-MfYPg0t4sj-FMLwRv1h:bgruyil) ([raw file](https://github.com/dairlab/dairlib/blob/fd188fcb7e87e52506a96dff7e1be66a072393c3/examples/Cassie/sim_cassie_sensor_aggregator.cc#L37)):*
> ```cpp
>       this->DeclareVectorInputPort("gyro", BasicVector<double>(3)).get_index();
>   radio_input_port_ =
>       this->DeclareAbstractInputPort("lcmt_radio_out Input Port",
> ```

I think this should just be radio
___
*[systems/controllers/linear_controller.cc, line 16 at r1](https://reviewable.io/reviews/dairlab/dairlib/250#-MfYRdiG697-8a5OtRBJ:-MfYRdiG697-8a5OtRBK:b-vjqgzc) ([raw file](https://github.com/dairlab/dairlib/blob/fd188fcb7e87e52506a96dff7e1be66a072393c3/systems/controllers/linear_controller.cc#L16)):*
> ```cpp
>   config_input_port_ =
>       this->DeclareVectorInputPort(
>               "config", LinearConfig(num_positions + num_velocities, num_inputs))
> ```

pd_config
___
*[systems/controllers/time_based_fsm.cc, line 29 at r1](https://reviewable.io/reviews/dairlab/dairlib/250#-MfYRlP4CJxZDdMqUgeb:-MfYRlP4CJxZDdMqUgec:b2ni2vv) ([raw file](https://github.com/dairlab/dairlib/blob/fd188fcb7e87e52506a96dff7e1be66a072393c3/systems/controllers/time_based_fsm.cc#L29)):*
> ```cpp
>           .get_index();
>   fsm_port_ = this->DeclareVectorOutputPort(
>                       "finite_state", BasicVector<double>(1),
> ```

fsm

Brian-Acosta:
I agree with not using `_in` and `_out` suffixes, I think diagrams will be easier to read that way

> I'm not sure about how to name ports that are of type `OutputVector`, robot_output feels vague but too wordy at the same time. We should probably distinguish between ports of type `BasicVector`, `TimestampedVector`, and `OutputVector` in the names, so that when we draw the diagram we can easily see which values are being passed along.

I think adding `(TimestampedVector)` to timestamped vector ports is probably sufficient here. Otherwise we can assume that the port is a `BasicVector` port unless it is named `robot_output` or whatever substitute for `robot_output` we decide on.

yangwill (commented):
I was thinking instead to name the vector-valued ports more succinctly like: 
- BasicVector for the robot's state: `x`
- TimestampedVector for the robot's state: `t, x`
- OutputVector for the robot's state and input: `t, x, u`
With the values in the same order as they appear in the vector.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/250#-:-MfcoGMu5FfVgh37eSO-:bz0vcsu)* status: all files reviewed, 4 unresolved discussions (waiting on @mposa)

Brian-Acosta (commented):
I like that too

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/250#-:-Mfcr1Zc8iTTJXGg9oGQ:br88gp4)* status: all files reviewed, 4 unresolved discussions (waiting on @mposa)

yminchen:
Just one comment, we should use `x_u_t` instead of `t_x_u` so that it's in the same order as how it's stored

yangwill (approved):
> Reviewed 22 of 22 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/250)* status: all files reviewed, 4 unresolved discussions (waiting on @mposa)

PR #258, opened 2021-09-30
Add default walking controller from goldilocks-model-dev branch

yminchen:


yangwill (changes requested):
> Reviewed 42 of 43 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 42 of 43 files reviewed, 29 unresolved discussions (waiting on @yangwill and @yminchen)

___
*[examples/Cassie/cassie_fixed_point_solver.cc, line 45 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mks5kNw844VpaMMwxbC:-Mks5kNw844VpaMMwxbD:b-llajr6) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/cassie_fixed_point_solver.cc#L45)):*
> ```cpp
>   auto left_toe = LeftToeFront(plant);
>   auto left_toe_evaluator = multibody::WorldPointEvaluator(plant,
>       left_toe.first, left_toe.second, rotationMatrix,
> ```

I'm not sure I understand why the toes use `rotationMatrix` and the heels use `ground_normal`. Also not sure why the naming convention (underscores/uppercase) is different for `ground_normal`. Can we make them consistent?
___
*[examples/Cassie/dispatcher_robot_out.cc, line 72 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksA1oX0n2l-K8AEsSj:-MksA1oX0n2l-K8AEsSk:b-u1gq6h) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/dispatcher_robot_out.cc#L72)):*
> ```cpp
>              " detected in which case it swtiches to test mode -1.");
> 
> // Initial condition (used for simulation)
> ```

Why is this disturbance located in dispatcher_out?
___
*[examples/Cassie/dispatcher_robot_out.cc, line 292 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAICyBkJPyWqi2zys:-MksAICyBkJPyWqi2zyt:bw5otpd) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/dispatcher_robot_out.cc#L292)):*
> ```cpp
>   builder.Connect(imu_passthrough->get_output_port(),
>                   robot_output_sender->get_input_port_imu());
>   if (FLAGS_broadcast_robot_state) {
> ```

Did we fix this TODO? We should just have one publish rate for the network and one publish rate for local - and have flags to set both publish rates.
___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-MksAqjr18HWWD7trsPg:bk39pjy) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*
> ```cpp
>   *v_result = result.GetSolution(v);
>   *u_result = result.GetSolution(u);
>   *lambda_result = result.GetSolution(lambda);
> ```

This way of setting the initial state should not be in `multibody_sim`. Also can you add a comment for why we need all these additional functions to set the initial state?
___
*[examples/Cassie/run_osc_walking_controller.cc, line 449 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktvJv_EDaPlTUOL0Uh:-MktvJv_EDaPlTUOL0Ui:bniq6ux) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/run_osc_walking_controller.cc#L449)):*
> ```cpp
> 
>   // Swing foot tracking
>   std::vector<double> swing_ft_ratio_breaks{0, left_support_duration / 2,
> ```

What is this section? Can we break it out somehow, either as gains or as a separate function similar to some cassie_utils functions?
___
*[examples/Cassie/visualizer.cc, line 62 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktvmnCC4fGsyuBWoTJ:-MktvmnCC4fGsyuBWoTK:b5u627j) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/visualizer.cc#L62)):*
> ```cpp
>   if (FLAGS_floating_base) {
>     // Ground direction
>     Eigen::Vector3d ground_normal(sin(FLAGS_ground_incline), 0,
> ```

I just realized that this is probably for your reduced order models stuff. We can bring it into master but we should probably be able to add in other terrain as well, which will probably be added in some other way. Maybe add a TODO to add varied terrain?
___
*[examples/Cassie/director_scripts/show_time.py, line 68 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktwQuF5ULTQdgy-ihs:-MktwQuG2Nf1pJPLof4V:b1qjaws) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/director_scripts/show_time.py#L68)):*
> ```Python
>             rt_ratio = dt / dt_real_time
> 
>             #my_text = my_text + ', real time factor: %.2f' % rt_ratio
> ```

We should just remove this line instead of commenting
___
*[examples/Cassie/osc/heading_traj_generator.cc, line 82 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktweK04uTkt_dR7h-m:-MktweK04uTkt_dR7h-n:bd5lic1) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/heading_traj_generator.cc#L82)):*
> ```cpp
>   /// Note that we construct trajectories in R^4 (quaternion space), so we need
>   /// to transform the yaw trajectory into quaternion representation.
>   double dt = 10;
> ```

Why is `dt` 10? We probably want some intuition for 10.
___
*[examples/Cassie/osc/high_level_command.cc, line 187 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mktx0Uz7zlWz_h_RTnz:-Mktx0Uz7zlWz_h_RTo-:b1kxeih) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/high_level_command.cc#L187)):*
> ```cpp
>   //////////// Get desired horizontal vel ////////////
>   // Calculate the current-desired yaw angle difference
>   double filtered_heading_error = (1 - weight) * 0 + weight * heading_error;
> ```

isn't this just `weight * heading_error`
___
*[examples/Cassie/osc/osc_walking_gains_mujoco.yaml, line 121 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktxA_l72sAycOG4Rsg:-MktxA_l72sAycOG4Rsh:b-eljg81) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/osc_walking_gains_mujoco.yaml#L121)):*
> ```YAML
>   [ 150,    0,     0,
>     0,  150,     0,
>     0,    0,   200]
> ```

What about MuJoCo makes us need a separate set of gains? Which ones work better on hardware?
___
*[examples/Cassie/osc/walking_speed_control.cc, line 134 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktxXDP8ZnOYpkkYOqI:-MktxXDP8ZnOYpkkYOqJ:bbovuk4) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/walking_speed_control.cc#L134)):*
> ```cpp
>   Vector3d pelvis_heading_vec =
>       plant_.EvalBodyPoseInWorld(*context_, pelvis_).rotation().col(0);
>   double approx_pelvis_yaw =
> ```

Why are we using this to calculate the pelvis heading? should that be directly in the robot's state? 
___
*[multibody/multibody_utils.cc, line 530 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mkty0BX1Au2f0F1NFpR:-Mkty0BX1Au2f0F1NFpS:b-iz737u) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/multibody/multibody_utils.cc#L530)):*
> ```cpp
> Eigen::MatrixXd JwrtqdotToJwrtv(
>     const Eigen::VectorXd& q, const Eigen::MatrixXd& Jwrtqdot) {
>   //[J_1:4, J_5:end] * [WToQuatDotMap, 0] = [J_1:4 * WToQuatDotMap, J_5:end]
> ```

Is this comment needed? It doesn't look like the final equation.
___
*[multibody/multibody_utils.cc, line 542 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktyVmd766YF0KZD4jG:-MktyVmd766YF0KZD4jH:bcf3tej) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/multibody/multibody_utils.cc#L542)):*
> ```cpp
> double GetTotalMass(const MultibodyPlant<double>& plant) {
>   // Get all body indices
>   std::vector<ModelInstanceIndex> model_instances;
> ```

This method exists in Drake as `CalcTotalMass`
___
*[systems/controllers/lipm_traj_gen.cc, line 183 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mktz_HK5UvosncYxku6:-Mktz_HK5UvosncYxku7:b-c46eti) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L183)):*
> ```cpp
>     double dist = toe_left_origin_position(1) - toe_right_origin_position(1);
>     // <0.2 meter: ratio 1
>     // >0.5 meter: ratio 0.8  //0.65
> ```

I think this comment is out of date. 
___
*[systems/controllers/lipm_traj_gen.cc, line 189 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzgAE1GJh8fSqz7cw:-MktzgAE1GJh8fSqz7cx:b-ev22i) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L189)):*
> ```cpp
>       heuristic_ratio_ = 0.9;
>     } else if (dist > foot_spread_lb_) {
>       heuristic_ratio_ = 1 + (0.9 - 1) / (foot_spread_ub_ - foot_spread_lb_) *
> ```

nitpick, it's hard to see that this is a linear interpolation, I think there are clearer ways to write a linear interpolation like `ratio = 0.8 + (1 - 0.8) * (dist - foot_spread_lb_)/(foot_spread_ub_ - foot_spread_lb_)
___
*[systems/controllers/lipm_traj_gen.cc, line 277 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzFIs-sIDzW7xr6To:-MktzFIs-sIDzW7xr6Tp:b1qhler) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L277)):*
> ```cpp
> }
> 
> void Print3dVector(VectorXd input) {
> ```

Do we need to keep this?
___
*[systems/controllers/lipm_traj_gen.cc, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku-4X025CRJyFhQYML:-Mku-4X025CRJyFhQYMM:btd3cno) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L354)):*
> ```cpp
>   *exp_pp_traj =
>       ConstructLipmTraj(CoM, dCoM, stance_foot_pos, start_time, end_time);
> ```

remove this comment
___
*[systems/controllers/swing_ft_traj_gen.cc, line 185 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0AQSAWmqps0MUrm7:-Mku0AQSAWmqps0MUrm8:b-b7flow) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L185)):*
> ```cpp
>       heuristic_ratio_ = 1;
>       if (dist > foot_spread_ub_) {
>         heuristic_ratio_ = ratio_lb_;
> ```

This heuristic is in both lipm and swing_ft. Maybe we can combine into one place?
___
*[systems/controllers/swing_ft_traj_gen.cc, line 260 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0Ggc4caVr5uw4uCC:-Mku0Ggc4caVr5uw4uCD:b-fj0ovu) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L260)):*
> ```cpp
>     for (int i = 0; i < CoM_vel_pred.size(); i++) {
>       outfile << CoM_vel_pred(i);
>       if (i == CoM_vel_pred.size() - 1) {
> ```

remove these comments.
___
*[systems/controllers/swing_ft_traj_gen.cc, line 314 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0IWb-2wA01b2ylgL:-Mku0IWb-2wA01b2ylgM:b-4q0oho) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L314)):*
> ```cpp
>     outfile << robot_output->get_timestamp() << ", ";
>     for (int i = 0; i < x_fs->size(); i++) {
>       outfile << (*x_fs)(i);
> ```

remove these comments as well
___
*[systems/controllers/time_based_fsm.cc, line 43 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0dG3AOnCYEQwSEjB:-Mku0dG3AOnCYEQwSEjC:b-r8y69v) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/time_based_fsm.cc#L43)):*
> ```cpp
>   for (auto& duration_i : state_durations) {
>     sum += duration_i;
>     accu_state_durations_.push_back(sum - eps_);
> ```

what is epsilon doing?
___
*[systems/controllers/time_based_fsm.cc, line 185 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0j2mA89HfnsKN4rc:-Mku0j2mA89HfnsKN4rd:bx9mq2k) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/time_based_fsm.cc#L185)):*
> ```cpp
>           state_port_))->get_timestamp();
> 
>   global_fsm_idx->get_mutable_value() << int((current_time - t0_ + 1e-8) /
> ```

what is global fsm? how is that different?
___
*[systems/controllers/osc/operational_space_control.h, line 347 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0xDEBuPWkV9vnqcC:-Mku0xDEBuPWkV9vnqcD:bacaipt) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L347)):*
> ```objc
>   // Testing contact force blend
>   double ds_duration_ = -1;
>   double w_blend_constraint_ = 0.1; //0.1 // for soft constraint
> ```

shouldn't this be the ds duration?
___
*[systems/controllers/osc/operational_space_control.h, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0zgd49yoJ6lb5d3W:-Mku0zge1Xz1qigoLUMH:b-rwp8n) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L354)):*
> ```objc
>   // Testing -- regularizing input
>   drake::solvers::QuadraticCost* input_reg_cost_;
>   double w_input_reg_ = 0.0000003;
> ```

we should just have it as an option in OSC
___
*[systems/controllers/osc/operational_space_control.cc, line 725 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1JwQCXU-KGgVDjck:-Mku1JwQCXU-KGgVDjcl:b-8qb00i) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L725)):*
> ```cpp
> 
>   //  const MathematicalProgramResult result = Solve(*prog_);
>   const MathematicalProgramResult result = solver_->Solve(*prog_);
> ```

Why are we using the slower solver?
___
*[systems/controllers/osc/osc_tracking_data.h, line 210 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1VYj3fN1ntpAhtKU:-Mku1VYj3fN1ntpAhtKV:bgpp7n1) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L210)):*
> ```objc
> 
>   // OscViewFrame
>   bool with_view_frame_ = false;
> ```

what is osc view frame?
___
*[systems/controllers/osc/osc_tracking_data.h, line 512 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1b7GE689tfD_eo9V:-Mku1b7GE689tfD_eo9W:b-ixlgsm) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L512)):*
> ```objc
> };
> 
> class TwoFrameTrackingData final : public OscTrackingData {
> ```

What does this class do?
___
*[systems/controllers/osc/osc_tracking_data.cc, line 66 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1kZoA9h3Fe6gxyOq:-Mku1kZoA9h3Fe6gxyOr:bgl1deh) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L66)):*
> ```cpp
>   if (track_at_current_state_) {
>     if (pre_update_) {
>       PreUpdate(x_w_spr, context_w_spr, x_wo_spr, context_wo_spr, traj, t,
> ```

what does preupdate do? can we just always use this?
___
*[systems/controllers/osc/osc_utils.h, line 18 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku22zn70gRBggLBjPp:-Mku22zn70gRBggLBjPq:bfwxqh5) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_utils.h#L18)):*
> ```objc
> // changes going on in the code from upstream
> Eigen::MatrixXd PositionMapFromSpringToNoSpring(
>     const drake::multibody::MultibodyPlant<double>& plant_w_spr,
> ```

this function already exists: `CreateWithSpringsToWithoutSpringsMapPos()` in `multibody_utils.h`

yangwill (commented):
I need some more time to look at osc_tracking_data. most of the comments are just to improve documentation.

One thing that I'm noticing is that we still have a lot of gains everywhere. Maybe we should have yamls for system specific parameters and have each system load its own parameters. I think ros works somewhat similarly.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258#-:-Mku2FCh96m0rw4CPxGs:bhakqln)* status: 42 of 43 files reviewed, 29 unresolved discussions (waiting on @yangwill and @yminchen)

Brian-Acosta (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 42 of 43 files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-Mkw83sU7-79-JYaZ_MI:bf8c8o4) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*

This way of setting the initial state should not be in `multibody_sim`. Also can you add a comment for why we need all these additional functions to set the initial state?
</blockquote></details>

I agree - can this be it's own function in cassie_utils or maybe even in a separate file? It seems like the purpose of this is to start Cassie with nonzero velocity and acceleration which I don't think we need to include in master
___
*[systems/controllers/BUILD.bazel, line 102 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwBn3JDJTg2FMpg06v:-MkwBn3JDJTg2FMpg06w:bne4d0y) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/BUILD.bazel#L102)):*
> ```Starlark
>     hdrs = ["time_based_fsm.h"],
>     deps = [
>         "//lcmtypes:lcmt_robot",
> ```

what is this lcmtype?
___
*[systems/controllers/osc/operational_space_control.h, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwDZk21icwYELMzaug:-MkwDZk21icwYELMzauh:bdeq6af) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L354)):*
> ```objc
>   // Testing -- regularizing input
>   drake::solvers::QuadraticCost* input_reg_cost_;
>   double w_input_reg_ = 0.0000003;
> ```

What is the purpose of w_input_reg? it seems really low - do we notice a difference if we remove it?
___
*[systems/controllers/osc/operational_space_control.cc, line 689 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwEG2hAMb87NtXKcW4:-MkwEG2hAMb87NtXKcW5:b-anz04f) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L689)):*
> ```cpp
>   if (ds_duration_ > 0) {
>     MatrixXd A = MatrixXd::Zero(1, 2 * n_c_ / kSpaceDim);
>     if ((fsm_state == 2) || (fsm_state == 3) || (fsm_state == 4)) {
> ```

We should think about the right way to not hard code these values. Perhaps it would be a good idea to define all of these things in the fsm header file?
___
*[systems/controllers/osc/operational_space_control.cc, line 975 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwFg_L2y_mH7Aowd15:-MkwFg_L2y_mH7Aowd16:b4alu9w) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L975)):*
> ```cpp
>   // Testing: Zero out the spring joints (avoid doing feedback with springs)
>   //  for(auto i : knee_ankle_pos_idx_list_) {q_w_spr(i) = 0;}
>   //  for(auto i : knee_ankle_vel_idx_list_) {v_w_spr(i) = 0;}
> ```

Remove or change to TODO so it gets tested
___
*[systems/controllers/osc/osc_tracking_data.h, line 134 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwG9GIDzANsc-ZaLwx:-MkwG9GIDzANsc-ZaLwy:b-odytzx) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L134)):*
> ```objc
>   bool pre_update_ = false;
> 
>   void SetLowPassFilter(double tau, const std::set<int>& element_idx = {});
> ```

Is inside the osc tracking data really the right place to be adding filters? and if it is, can we use the InternalLowpassFilter class to keep things tidy?
___
*[systems/controllers/osc/osc_tracking_data.cc, line 245 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwHZOZBGR6UMyLklZO:-MkwHZO_2DbR1nO6FOLn:b-tle97s) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L245)):*
> ```cpp
> // Run this function in OSC constructor to make sure that users constructed
> // OscTrackingData correctly.
> void OscTrackingData::CheckOscTrackingData(bool no_control_gains) {
> ```

What does this flag mean?
___
*[systems/controllers/osc/osc_tracking_data.cc, line 661 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwIDyL-9IUA5siLsQK:-MkwIDyL-9IUA5siLsQL:b21tw73) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L661)):*
> ```cpp
> 
> /**** TwoFrameTrackingData ****/
> TwoFrameTrackingData::TwoFrameTrackingData(
> ```

The way this is implemented (specifically how GetY() is formulated) makes this seem like it is written to be for tracking relative positions of body frames -  I don't think this would be acccurate for relative rotations. Can we change the name and constructor to specify that the child tracking datas are Translational? i.e. RelativeTranslationTrackingData
___
*[systems/controllers/osc/osc_tracking_data.cc, line 666 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwJ3BA05GxdKZaLud4:-MkwJ3BA05GxdKZaLud5:b48bbre) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L666)):*
> ```cpp
>     const drake::multibody::MultibodyPlant<double>& plant_w_spr,
>     const drake::multibody::MultibodyPlant<double>& plant_wo_spr,
>     OscTrackingData& to_frame, OscTrackingData& from_frame)
> ```

Can we call this to_frame_data and from_frame_data ?

Brian-Acosta (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 42 of 43 files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

yminchen (commented):
@yangwill @brian-acosta Thanks for the reviews. I've addressed a few comments. Will continue soon.

> Reviewed 1 of 43 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258#-:-MkwdYB0AvTK9X8vA1aO:b4hz7xf)* status: 37 of 43 files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[examples/Cassie/cassie_fixed_point_solver.cc, line 45 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mks5kNw844VpaMMwxbC:-Mkw9hJCCLzJoesrh2Eq:bb6un1j) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/cassie_fixed_point_solver.cc#L45)):*

I'm not sure I understand why the toes use `rotationMatrix` and the heels use `ground_normal`. Also not sure why the naming convention (underscores/uppercase) is different for `ground_normal`. Can we make them consistent?
</blockquote></details>

Done
___
*[examples/Cassie/dispatcher_robot_out.cc, line 72 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksA1oX0n2l-K8AEsSj:-MkwXaktBqC9ckIF7fUR:b-7863aq) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/dispatcher_robot_out.cc#L72)):*

Why is this disturbance located in dispatcher_out?
</blockquote></details>

This is to initialize the state estimator when we test controllers in simulation. 
When we set an initial velocity in the simulation, we also need to initialize the state estimator with that velocity.
___
*[examples/Cassie/dispatcher_robot_out.cc, line 292 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAICyBkJPyWqi2zys:-MkwYwqT-J-yX8stztvU:byvyago) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/dispatcher_robot_out.cc#L292)):*

Did we fix this TODO? We should just have one publish rate for the network and one publish rate for local - and have flags to set both publish rates.
</blockquote></details>

Yes. Done.
___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-MkwbSBo2T6laS1wyrfw:b-ibmo50) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

I agree - can this be it's own function in cassie_utils or maybe even in a separate file? It seems like the purpose of this is to start Cassie with nonzero velocity and acceleration which I don't think we need to include in master
</blockquote></details>

Yes, I can move this to cassie util or a separate file. 
We need `VdotConstraint` so it's dynamically feasible. As for `BodyPointVelConstraint`, we might have a constraint available in dircon that I can replace with. 

@brian-acosta @yangwill I can remove this pelvis velocity initialization if you guys think you won't use it.
___
*[systems/controllers/osc/osc_tracking_data.cc, line 66 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1kZoA9h3Fe6gxyOq:-MkwFuek5QltwKQem0OE:bte79b4) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L66)):*

what does preupdate do? can we just always use this?
</blockquote></details>

It updates the kinematics data for the `OscTrackingData`'s in `TwoFrameTrackingData`. E.g. we have to compute the positions of both frames before we can get the relative position.
This is the first implementation that came to mind. Maybe there is a better implementation. 
___
*[systems/controllers/osc/osc_tracking_data.cc, line 245 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwHZOZBGR6UMyLklZO:-MkwVcAT2S9XT3vXt1hN:bze5eyn) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L245)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

What does this flag mean?
</blockquote></details>

When the flag is on, it doesn't check the dimensions of Kp, Kd and W. 
We use this for the `TrackingData`'s in TwoFrameTrackingData, since we only need gains for the relative position.
___
*[systems/controllers/osc/osc_tracking_data.cc, line 661 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwIDyL-9IUA5siLsQK:-MkwNxzM8BMcGpXHjlrc:b-8c716j) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L661)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

The way this is implemented (specifically how GetY() is formulated) makes this seem like it is written to be for tracking relative positions of body frames -  I don't think this would be acccurate for relative rotations. Can we change the name and constructor to specify that the child tracking datas are Translational? i.e. RelativeTranslationTrackingData
</blockquote></details>

You are right. I wrote this class for relative transnational position. I will rename it to RelativeTranslationTrackingData.
___
*[systems/controllers/osc/osc_tracking_data.cc, line 666 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwJ3BA05GxdKZaLud4:-MkwXPhUE04tlHCoKmza:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.cc#L666)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Can we call this to_frame_data and from_frame_data ?
</blockquote></details>

Done.

Brian-Acosta (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 37 of 43 files reviewed, 34 unresolved discussions (waiting on @yangwill and @yminchen)

___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-Mkx2igk5q6ogi1UUwkH:bptrupy) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*

Yes, I can move this to cassie util or a separate file. 
We need `VdotConstraint` so it's dynamically feasible. As for `BodyPointVelConstraint`, we might have a constraint available in dircon that I can replace with. 

@brian-acosta @yangwill I can remove this pelvis velocity initialization if you guys think you won't use it.
</blockquote></details>

I removed it on my local copy of this branch - I will commit these changes

yminchen (commented):
> Reviewed 5 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 42 of 43 files reviewed, 34 unresolved discussions (waiting on @Brian-Acosta and @yangwill)

___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-MkxhJ854XV82KoNVN1o:bykled8) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

I removed it on my local copy of this branch - I will commit these changes
</blockquote></details>

I can also just push my code here if it's fine.
___
*[examples/Cassie/run_osc_walking_controller.cc, line 449 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktvJv_EDaPlTUOL0Uh:-Mkx0TV07aLuGacQoyR0:bq4vnqh) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/run_osc_walking_controller.cc#L449)):*

What is this section? Can we break it out somehow, either as gains or as a separate function similar to some cassie_utils functions?
</blockquote></details>

The ratios here set the profile (time varying) for the Kp Kd gains. I am renaming `ratio` to `gain_multiplier` to be more accurate.

I was thinking of moving this into yaml file, but the time breaks are functions of `left_support_duration`. I also don't think this belongs to Cassie utils, because this is specific to walking. Maybe just leave this here for now unless we have a good solution?  
___
*[examples/Cassie/visualizer.cc, line 62 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktvmnCC4fGsyuBWoTJ:-Mkx2r4b0amGdWzfLmfN:b-7yd1q4) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/visualizer.cc#L62)):*

I just realized that this is probably for your reduced order models stuff. We can bring it into master but we should probably be able to add in other terrain as well, which will probably be added in some other way. Maybe add a TODO to add varied terrain?
</blockquote></details>

Added a TODO.
___
*[examples/Cassie/director_scripts/show_time.py, line 68 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktwQuF5ULTQdgy-ihs:-Mkx3w6lCGHB_0iKv8RZ:b-5wmgva) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/director_scripts/show_time.py#L68)):*

We should just remove this line instead of commenting
</blockquote></details>

I still use realtime rate from time to time. I will uncomment it back
___
*[examples/Cassie/osc/heading_traj_generator.cc, line 82 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktweK04uTkt_dR7h-m:-Mkx4Q8kCklth1HnxGt8:b-87i3f7) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/heading_traj_generator.cc#L82)):*

Why is `dt` 10? We probably want some intuition for 10.
</blockquote></details>

This number changes the the trajectory time horizon. As long as it's larger than the recompute time, the value doesn't affect the control outcome. 
Added this to documentation.
___
*[examples/Cassie/osc/high_level_command.cc, line 187 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mktx0Uz7zlWz_h_RTnz:-Mkx5uD0AHWDOjS9UmNY:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/high_level_command.cc#L187)):*

isn't this just `weight * heading_error`
</blockquote></details>

Done.
___
*[examples/Cassie/osc/osc_walking_gains_mujoco.yaml, line 121 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktxA_l72sAycOG4Rsg:-Mkx79f86WRyvbGD6FlL:b-yfs8hk) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/osc_walking_gains_mujoco.yaml#L121)):*

What about MuJoCo makes us need a separate set of gains? Which ones work better on hardware?
</blockquote></details>

I created a copy of gains for mujoco because I have been tuning gains in mujoco or simply played with the controller in mujoco. And I didn't want to accidentally use these untested gains on Cassie by accident.

The file without `_mujoco` affix are for the hardware. The optimal gains on hardware are different from those in mujoco.
___
*[examples/Cassie/osc/walking_speed_control.cc, line 134 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktxXDP8ZnOYpkkYOqI:-MkxAQGL0or9cXqEDzV_:b-yx5en7) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/osc/walking_speed_control.cc#L134)):*

Why are we using this to calculate the pelvis heading? should that be directly in the robot's state? 
</blockquote></details>

I only want to rotate about the global z axis, but the quaternion in robot's state includes roll and pitch. 
___
*[multibody/multibody_utils.cc, line 530 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mkty0BX1Au2f0F1NFpR:-MkxCclj-yUNd6guPasr:bsf5g8s) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/multibody/multibody_utils.cc#L530)):*

Is this comment needed? It doesn't look like the final equation.
</blockquote></details>

Yeah, this is the documentation for this function. The right hand side of the equation is `ret` four lines below here. 
___
*[multibody/multibody_utils.cc, line 542 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktyVmd766YF0KZD4jG:-MkxDuG-BYuKbVWz-eE0:b-x2xena) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/multibody/multibody_utils.cc#L542)):*

This method exists in Drake as `CalcTotalMass`
</blockquote></details>

Good catch. What is a fossil...
___
*[systems/controllers/BUILD.bazel, line 102 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwBn3JDJTg2FMpg06v:-MkxF1YX-43CYH-Tsi-V:byv47e8) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/BUILD.bazel#L102)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

what is this lcmtype?
</blockquote></details>

It is a collection of all lcmtypes: https://github.com/DAIRLab/dairlib/blob/master/lcmtypes/BUILD.bazel#L19
___
*[systems/controllers/lipm_traj_gen.cc, line 183 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mktz_HK5UvosncYxku6:-MkxHF39AzRUuVDfak4V:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L183)):*

I think this comment is out of date. 
</blockquote></details>

Done.
___
*[systems/controllers/lipm_traj_gen.cc, line 189 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzgAE1GJh8fSqz7cw:-MkxHFId7OaPqqwjxMLJ:b-47g3xh) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L189)):*

nitpick, it's hard to see that this is a linear interpolation, I think there are clearer ways to write a linear interpolation like `ratio = 0.8 + (1 - 0.8) * (dist - foot_spread_lb_)/(foot_spread_ub_ - foot_spread_lb_)
</blockquote></details>

I think there is a mistake in the equation you show. `dist` = foot_spread_ub_ should gives 0.9 not 1
___
*[systems/controllers/lipm_traj_gen.cc, line 277 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzFIs-sIDzW7xr6To:-MkxGgt770zzY6STFHp5:bss4a2b) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L277)):*

Do we need to keep this?
</blockquote></details>

Removed.
___
*[systems/controllers/lipm_traj_gen.cc, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku-4X025CRJyFhQYML:-MkxGiD_2XAHWlF3JaUW:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L354)):*

remove this comment
</blockquote></details>

Done.
___
*[systems/controllers/swing_ft_traj_gen.cc, line 185 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0AQSAWmqps0MUrm7:-MkxIdZd66tXsgJ9NLmd:bz3zie8) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L185)):*

This heuristic is in both lipm and swing_ft. Maybe we can combine into one place?
</blockquote></details>

We probably only need one of them, but I haven't tested them thoroughly yet. I'm adding a TODO to this 
___
*[systems/controllers/swing_ft_traj_gen.cc, line 260 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0Ggc4caVr5uw4uCC:-MkxK3Z97iy2ezTlj-aV:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L260)):*

remove these comments.
</blockquote></details>

Done.
___
*[systems/controllers/swing_ft_traj_gen.cc, line 314 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0IWb-2wA01b2ylgL:-MkxLB3z97nCGajyytql:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/swing_ft_traj_gen.cc#L314)):*

remove these comments as well
</blockquote></details>

Done.
___
*[systems/controllers/time_based_fsm.cc, line 43 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0dG3AOnCYEQwSEjB:-MkxORnk0Tv7Kxi1Zsc1:b-i54d44) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/time_based_fsm.cc#L43)):*

what is epsilon doing?
</blockquote></details>

There was a floating point precision issue in comparing two floating point numbers (I think it's `remainder < accu_state_durations_[i]`). FSM didn't get to the next state when it's supposed to, which my MPC is very sensitive to. 
Should I add a flag for this eps margin? 
___
*[systems/controllers/time_based_fsm.cc, line 185 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0j2mA89HfnsKN4rc:-MkxLZJ99VGGhX1OdTBs:b-xggu4) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/time_based_fsm.cc#L185)):*

what is global fsm? how is that different?
</blockquote></details>

global_fsm_idx tells us how many number of fsm period we have been through. It's used in my reduced order model MPC. 
___
*[systems/controllers/osc/operational_space_control.h, line 347 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0xDEBuPWkV9vnqcC:-MkxScGw3iMJ7ZDFFcb3:b-92cz6m) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L347)):*

shouldn't this be the ds duration?
</blockquote></details>

`w_blend_constraint` is the soft constraint weight if we decides to use make the constraint soft
___
*[systems/controllers/osc/operational_space_control.h, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwDZk21icwYELMzaug:-MkxREsnC3fFWcMCjjK_:b6m020l) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L354)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

What is the purpose of w_input_reg? it seems really low - do we notice a difference if we remove it?
</blockquote></details>

Yes, the OSC output is very sensitive to this parameter, since u is ~10^2
___
*[systems/controllers/osc/operational_space_control.h, line 354 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku0zgd49yoJ6lb5d3W:-MkxT5-5EORu8ZaPamqb:bltcvl9) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.h#L354)):*

we should just have it as an option in OSC
</blockquote></details>

I will add this to yaml file.
___
*[systems/controllers/osc/operational_space_control.cc, line 689 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwEG2hAMb87NtXKcW4:-MkxXkzCCUACWe5JdIwC:b-q3l9fj) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L689)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

We should think about the right way to not hard code these values. Perhaps it would be a good idea to define all of these things in the fsm header file?
</blockquote></details>

I added a function that set the states for blending.
___
*[systems/controllers/osc/operational_space_control.cc, line 725 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1JwQCXU-KGgVDjck:-Mkx_uo7EYDBEgcij9YX:bs4a26c) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L725)):*

Why are we using the slower solver?
</blockquote></details>

It's using the fast qp solver. I have removed the dead code.
___
*[systems/controllers/osc/operational_space_control.cc, line 975 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwFg_L2y_mH7Aowd15:-Mkxa6Ko6WSc95hvF8kT:b3lz7o8) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/operational_space_control.cc#L975)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Remove or change to TODO so it gets tested
</blockquote></details>

Missed this when deleting the test code
___
*[systems/controllers/osc/osc_tracking_data.h, line 134 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MkwG9GIDzANsc-ZaLwx:-MkxaO-MAm4xw2TBwQTY:b-7pfkmd) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L134)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Is inside the osc tracking data really the right place to be adding filters? and if it is, can we use the InternalLowpassFilter class to keep things tidy?
</blockquote></details>

I've been wanting to use your low pass filter for a while, but it's not merged into master yet!
___
*[systems/controllers/osc/osc_tracking_data.h, line 210 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1VYj3fN1ntpAhtKU:-MkxaZz47l3RRQ8Jlz7A:bmhlm1q) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L210)):*

what is osc view frame?
</blockquote></details>

I added this to rotate the frame for translational task space tracking data. One use case: we can set the swing foot tracking gains in the local frame instead of the global frame.
___
*[systems/controllers/osc/osc_tracking_data.h, line 512 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku1b7GE689tfD_eo9V:-Mkxdom12j-p1fOkSt3x:b-xpnkt1) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_tracking_data.h#L512)):*

What does this class do?
</blockquote></details>

I just added documentation for this class
___
*[systems/controllers/osc/osc_utils.h, line 18 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-Mku22zn70gRBggLBjPp:-MkxeTfCD-xmkjBskaiP:b1hfu1u) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/osc/osc_utils.h#L18)):*

this function already exists: `CreateWithSpringsToWithoutSpringsMapPos()` in `multibody_utils.h`
</blockquote></details>

I found that `CreateWithSpringsToWithoutSpringsMapPos` and `CreateWithSpringsToWithoutSpringsMapVel` don't do what the names suggest. Currently,  `CreateWithSpringsToWithoutSpringsMapPos` maps `state_w_spring` to `state_wo_spring`, and `CreateWithSpringsToWithoutSpringsMapVel` does the inverse map.

I can fix the bugs in this PR if you are fine with it

Brian-Acosta (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 29 of 44 files reviewed, 34 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-MkxioQtD0We7iJuj8lN:b-lnrp1f) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*

I can also just push my code here if it's fine.
</blockquote></details>

Oh I already pushed lol

yminchen (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: 29 of 44 files reviewed, 34 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[examples/Cassie/multibody_sim.cc, line 617 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MksAqjr18HWWD7trsPf:-MkxjPVX6AtFHy3Cb6ok:bblx4m5) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/examples/Cassie/multibody_sim.cc#L617)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Oh I already pushed lol 
</blockquote></details>

Yeah just saw it. I can revert it back now

yminchen (commented):
Done addressing all comments so far 

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258#-:-MkxjiKe4pMeL8cp7PK2:baxbzfd)* status: 29 of 44 files reviewed, 34 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

Brian-Acosta (changes requested):
> Reviewed 10 of 14 files at r4, 1 of 1 files at r5, 6 of 6 files at r6, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258)* status: all files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[multibody/multibody_utils.h, line 147 at r6](https://reviewable.io/reviews/dairlab/dairlib/258#-Ml88Jxh-MxdsyOKzIPz:-Ml88Jxh-MxdsyOKzIQ-:bn8exfv) ([raw file](https://github.com/dairlab/dairlib/blob/d7ad189fdbe380d69682e3849653586b6d8bcc57/multibody/multibody_utils.h#L147)):*
> ```objc
> /// Ref: equation 16 of https://arxiv.org/pdf/0811.2889.pdf
> /// Note: The same calculation exits in Drake's
> /// QuaternionFloatingMobilizer<T>::AngularVelocityToQuaternionRateMatrix()
> ```

Is this in drake's internal namespace? If not can we just use it and if it is can we ask to expose it?
___
*[systems/controllers/lipm_traj_gen.cc, line 189 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzgAE1GJh8fSqz7cw:-Ml88uctCKPii-85xqeV:beqj93v) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L189)):*

I think there is a mistake in the equation you show. `dist` = foot_spread_ub_ should gives 0.9 not 1
</blockquote></details>

It's fine how it is but I think this could be pretty clear as a one liner if we wanted:
heuristic_ratio_ = drake::math::saturate(1 +  (0.9-1)*(dist - foot_spread_lb_) / (foot_spread_ub_ - foot_spread_lb_), 0.9, 1.0);
___
*[systems/controllers/osc/osc_tracking_data.h, line 21 at r6](https://reviewable.io/reviews/dairlab/dairlib/258#-Ml8CkGaBuVm1hClQrTP:-Ml8CkGaBuVm1hClQrTQ:boq3p1v) ([raw file](https://github.com/dairlab/dairlib/blob/d7ad189fdbe380d69682e3849653586b6d8bcc57/systems/controllers/osc/osc_tracking_data.h#L21)):*
> ```objc
> /// One use case: we can set the swing foot tracking gains in the local frame
> /// instead of the global frame.
> class OscViewFrame {
> ```

I think this is a really good idea and I'm not sure how we didn't have something like this already. Could we make the name a little more clear? Maybe LocalTrackingDataFrame or BodyFixedTrackingDataFrame? 
___
*[systems/controllers/osc/osc_tracking_data.cc, line 262 at r6](https://reviewable.io/reviews/dairlab/dairlib/258#-Ml8ECtv53CroTQOfZKc:-Ml8ECtv53CroTQOfZKd:bwbyc4q) ([raw file](https://github.com/dairlab/dairlib/blob/d7ad189fdbe380d69682e3849653586b6d8bcc57/systems/controllers/osc/osc_tracking_data.cc#L262)):*
> ```cpp
> }
> 
> void OscTrackingData::UpdateJAndJdotVForUnitTest(
> ```

Can we move this somewehere else? seems like a weird place for this

yminchen (commented):
Addressed all comments so far

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258#-:-MlBgtbS6DgoiPfG-HC2:ba1hrou)* status: 43 of 44 files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

___
*[multibody/multibody_utils.h, line 147 at r6](https://reviewable.io/reviews/dairlab/dairlib/258#-Ml88Jxh-MxdsyOKzIPz:-MlBOk6o-581kUhfDgEg:bgo0vxw) ([raw file](https://github.com/dairlab/dairlib/blob/d7ad189fdbe380d69682e3849653586b6d8bcc57/multibody/multibody_utils.h#L147)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Is this in drake's internal namespace? If not can we just use it and if it is can we ask to expose it?
</blockquote></details>

Yes it's internal drake's function. 

Per in-person discussion, we chose to have our own function, since Drake's `QuaternionFloatingMobilizer` class has many other methods,  and Drake team will probably not expose only one method for us.
___
*[systems/controllers/lipm_traj_gen.cc, line 189 at r1](https://reviewable.io/reviews/dairlab/dairlib/258#-MktzgAE1GJh8fSqz7cw:-MlBRAoID9Hj_L3L_wPx:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/c7b6682ab79c7e32d1a92a54bfb9daae121085fd/systems/controllers/lipm_traj_gen.cc#L189)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

It's fine how it is but I think this could be pretty clear as a one liner if we wanted:
heuristic_ratio_ = drake::math::saturate(1 +  (0.9-1)*(dist - foot_spread_lb_) / (foot_spread_ub_ - foot_spread_lb_), 0.9, 1.0);
</blockquote></details>

Done.
___
*[systems/controllers/osc/osc_tracking_data.cc, line 262 at r6](https://reviewable.io/reviews/dairlab/dairlib/258#-Ml8ECtv53CroTQOfZKc:-MlBatBL3nCveXGADbpE:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/d7ad189fdbe380d69682e3849653586b6d8bcc57/systems/controllers/osc/osc_tracking_data.cc#L262)):*
<details><summary><i>Previously, Brian-Acosta (Brian Acosta) wrote…</i></summary><blockquote>

Can we move this somewehere else? seems like a weird place for this
</blockquote></details>

Done.

yminchen (commented):
One TODO is the naming of `OscViewFrame`

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/258#-:-MlBh2p9AamaNHv3Zbko:b-jovf8u)* status: 43 of 44 files reviewed, 37 unresolved discussions (waiting on @Brian-Acosta, @yangwill, and @yminchen)

PR #270, opened 2021-11-16
Fix a bug in pelvis balance traj generation

yminchen:
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/270)


yminchen:
The final target quaternion for the pitch and roll was set incorrectly. They should be 0.

Brian-Acosta (approved):
LGTM

PR #283, opened 2022-02-25
Add Geom-Geom collider for basic queries

mposa:
Adds a simple class to compute signed distances between geometries and their Jacobians, including a polytope representation of the frictional Jacobian. If useful, it would also be easy to add a variation which doesn't compute the polytope.

Drake has this existing functionality scattered about, but not all in one place.

---
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/283)

### DIFF
--- a/examples/trifinger/simulate_trifinger.py
+++ b/examples/trifinger/simulate_trifinger.py
@@ -9,7 +9,7 @@ class TrifingerDemoController(LeafSystem):
     LeafSystem.__init__(self)
 
     self.plant = plant
-        
+
     # Input is state, output is torque (control action)
     self.DeclareVectorInputPort("x", plant.num_positions() +
         plant.num_velocities())
--- a/multibody/BUILD.bazel
+++ b/multibody/BUILD.bazel
@@ -102,3 +102,24 @@ cc_library(
     name = "ball_urdf",
     data = glob(["multibody/ball.urdf"]),
 )
+
+cc_library(
+    name = "geom_geom_collider",
+    srcs = ["geom_geom_collider.cc"],
+    hdrs = ["geom_geom_collider.h"],
+    deps = [
+        "@drake//:drake_shared_library",
+    ]
+)
+
+cc_binary(
+    name = "geom_geom_collider_test",
+    srcs = ["test/geom_geom_collider_test.cc"],
+    deps = [
+        ":geom_geom_collider",
+        ":utils",
+        "//common",
+        "@drake//:drake_shared_library",
+    ],
+    data = ["//examples/trifinger:trifinger_urdf"],
+)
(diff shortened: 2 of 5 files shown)


mshalm (commented):
> Reviewed 5 of 5 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/283)* status: all files reviewed, 4 unresolved discussions

___
*[multibody/geom_geom_collider.h, line 30 at r1](https://reviewable.io/reviews/dairlab/dairlib/283#-MwmZy1a7z50H3Ri4cE2:-MwmZy1a7z50H3Ri4cE3:bozzmpf) ([raw file](https://github.com/dairlab/dairlib/blob/2b6e5fbbfacd5ae71f2426ee195424a501b4fca2/multibody/geom_geom_collider.h#L30)):*
> ```objc
>       const drake::geometry::GeometryId geometry_id_A,
>       const drake::geometry::GeometryId geometry_id_B,
>       const int num_friction_directions);
> ```

I would suggest at least providing an alternative option where the frictional jacobian is just the jacobian of the contact-frame velocity (without projecting onto N basis vectors)
___
*[multibody/geom_geom_collider.h, line 50 at r1](https://reviewable.io/reviews/dairlab/dairlib/283#-MwmaxeIBGJMOlHzhnJh:-MwmallO-Bn5LkTV8Ep2:bonxde3) ([raw file](https://github.com/dairlab/dairlib/blob/2b6e5fbbfacd5ae71f2426ee195424a501b4fca2/multibody/geom_geom_collider.h#L50)):*
> ```objc
>   /// Jacobian is ordered [J_n; J_t], and has shape
>   /// (1 + 2 * num_friction_directions_) x (nq)
>   /// @param context
> ```

Maybe specify that `context` is `plant_`'s Context, at least in the comment?
___
*[multibody/geom_geom_collider.h, line 51 at r1](https://reviewable.io/reviews/dairlab/dairlib/283#-MwmdT1j51jmj-aCmPPO:-MwmdT1j51jmj-aCmPPP:bqz7owz) ([raw file](https://github.com/dairlab/dairlib/blob/2b6e5fbbfacd5ae71f2426ee195424a501b4fca2/multibody/geom_geom_collider.h#L51)):*
> ```objc
>   /// (1 + 2 * num_friction_directions_) x (nq)
>   /// @param context
>   /// @param J
> ```

comment should specify the variable that `J` is wrt-- currently qDot, but am also wondering why that is a default instead of v. Adding the w.r.t. variable as a parameter of `GeomGeomCollider` would also be a reasonable option here.
___
*[multibody/geom_geom_collider.cc, line 28 at r1](https://reviewable.io/reviews/dairlab/dairlib/283#-MwmbLze22Xu0Yovw2HW:-MwmbLze22Xu0Yovw2HX:b-hp391w) ([raw file](https://github.com/dairlab/dairlib/blob/2b6e5fbbfacd5ae71f2426ee195424a501b4fca2/multibody/geom_geom_collider.cc#L28)):*
> ```cpp
>       num_friction_directions_(num_friction_directions) ,
>       planar_normal_(Vector3d::Zero()) {
>   if (num_friction_directions == 1) {
> ```

`<= 1`?

PR #304, opened 2022-06-07
ros update

Brian-Acosta:
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/304)


Brian-Acosta:
ROS update - should set up ros 20.04 CI before merging

PR #319, opened 2022-10-06
python script for visualizing offline generated trajectories

yangwill:
Visualization script similar to `visualize_trajectory.cc` written in Python for easier modification.

Supports visualization in both MeshCat and drake_visualizer. Meshcat support added to make additional visualization easier and to slowly deprecate drake_visualizer.

added `.yaml` to take the place of gflags to save visualization configurations and make modifications easier.


yminchen (changes requested):
Only skimmed through the codes. Could you
* provide an step to step instruction (maybe a readme file in `bindings/pydairlib/lcm/visualization`?) for how to use the script? I will go through it to visualize it on my computer
* remove any dead code if it exists

PR #324, opened 2022-11-01
Enable complex mpc walk

ShaneRozenLevy:
Ideally should be multiple PRs, but this PR:

-  Fix MPC formulation to treat momentum as a decision variable
-  Changed to trapazoidal collocation
-  Provided interface for setting mode sequence
-  Added constraint for initial state
-  Added interface for setting initial guess for all decision variables
-  Created gain struct and associated yaml
-  Created reference generation class with creates a references for walking/running gait using gait specification, nominal stand, and com velocity keyframes.

Per file changes, broken into the 3 main goals for this PR:
**MPC Formulation**
`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.*`: Fixed mpc formulation, uses mode sequence, and interface changes to use other changes
`multibody/kinematic/kinematic_constraints.cc`: Added a negation to properly constraint world point evaluator to foot position decision variable
`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.*`: Reworked constraint for fixed mpc formulation
`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc`: heavy rework given new code

**Quality of life**
`examples/Cassie/cassie_utils.*`: Add helper function for getting joint names
`examples/Cassie/kinematic_centroidal_mpc/cassie_kinematic_centroidal_mpc.*`: Cassie specific child class of `kinematic_centroidal_mpc` for reduced copies code in for examples
`examples/Cassie/kinematic_centroidal_mpc/kinematic_centroidal_mpc_gains.yaml`: Gains yaml
`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_gains.h`: Gains class

**Reference generation**
`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.*`: Class for generating simple reference trajectory
`systems/controllers/kinematic_centroidal_mpc/gait.*`: Class used for reference generation gait specification
`examples/Cassie/kinematic_centroidal_mpc/cassie_reference_utils.*`: Broke out IK code for nominal stand to its own file
`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.*` Utility file for reference generation class

- - -
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324)


yangwill (changes requested):
Midway through review - looks good, mostly nitpicks so far

> Reviewed 7 of 22 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324#-:-NFzMptk7kW8DDRqYT0z:b-2ot8z1)* status: 7 of 22 files reviewed, 7 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[`examples/Cassie/cassie_utils.cc` line 244 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzI6Cq4J75YDDuI1GH:-NFzI6Cq4J75YDDuI1GI:b-i44l3n) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/cassie_utils.cc#L244)):*
> ```cpp
>   };
>   std::vector<std::string> sym_joint_names{"hip_pitch", "knee", "ankle_joint", "toe"};
>   std::vector<std::string> joint_names{};
> ```

We've separate the joints into symmetric vs asymmetric in order to correctly implement periodicity constraints. Given that we're just pushing them all to the same vector, we might want to either not distinguish between the two types OR return them as separate vectors that users can then concatenate.
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc` line 50 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzInqA1-4cVU8VaAjU:-NFzInqA1-4cVU8VaAjV:b-emfm4y) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc#L50)):*
> ```cpp
> 
>   // Create gaits
>   Gait stand ({{0, 1, drake::Vector<bool, 4>(true, true, true, true)}});
> ```

I like that you've moved the settings into a yaml, can we also move these gait definitions into separate yamls? Kind of like a stand/walk yaml
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc` line 121 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzJF0jBretK4gsvS_k:-NFzJF0jBretK4gsvS_l:b-nfd2uu) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc#L121)):*
> ```cpp
>   std::cout << "Solving optimization\n\n";
>   const auto pp_xtraj = mpc.Solve();
>   mpc.SaveSolutionToFile(std::string(getenv("HOME")) + "/workspace/dairlib/examples/Cassie/saved_trajectories/kcmpc_solution");
> ```

although we try to keep our file structure consistent, it'd be nice to keep it all within the repo - which is best accomplished using paths relative to the repo. Generally, I've done this by always running binaries from the root of the repo - but open to other suggestions.
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kinematic_centroidal_mpc.h` line 4 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzK59m4-ogYm4svDKY:-NFzK59m4-ogYm4svDKZ:b-4a7j78) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kinematic_centroidal_mpc.h#L4)):*
> ```objc
> #include <iostream>
> #include "systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.h"
> #include "external/drake/tools/install/libdrake/_virtual_includes/drake_shared_library/drake/multibody/plant/multibody_plant.h"
> ```

nitpick, clion sometimes does this when it suggests imports but the correct path should be #include <drake/multibody/plant/multibody_plant.h>
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.h` line 33 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzKjMc6lHaDBtBev9y:-NFzKjMc6lHaDBtBev9z:b-ieafdu) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.h#L33)):*
> ```objc
>    * @brief Checks to make sure the gait is valid, Asserts if not valid
>    */
>   void Is_Valid() const;
> ```

i think drake style guide prefers `is_valid()`.
Though because this does not return anything, I think it should be `check_valid()`.
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.h` line 43 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzL2NE6Cbj2y5m883j:-NFzL2NE6Cbj2y5m883k:b-wnaqft) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.h#L43)):*
> ```objc
>    * @return
>    */
>   int CurrentMode(double time_now) const;
> ```

Also weird style preference of drake. I think trivial getters are `lower_case()` and "verbs" or functions that compute things are `GetUpperCase()`.
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.cc` line 7 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzLsm6CqxHKZpnuhh8:-NFzLsm73drx182foyay:b-vph5a8) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.cc#L7)):*
> ```cpp
>   for(int i = 0; i < gait_pattern_.size(); i++){
>     const auto& mode = gait_pattern_[i];
>     if(mode.start_phase <= phase_now && mode.end_phase > phase_now){
> ```

If we're returning immediately when we find a match, and if mode.start_phase is monotonically increasing, we should only have to check `phase_now > mode.start_phase`. Though considering the second condition (which should always be true if the first condition is true in the order that we check) is after, this optimization is negligible.

ShaneRozenLevy:
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc` line 121 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzJF0jBretK4gsvS_k:-NFzOBqvATdG9piH0iZs:bky40u6) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc#L121)):*

although we try to keep our file structure consistent, it'd be nice to keep it all within the repo - which is best accomplished using paths relative to the repo. Generally, I've done this by always running binaries from the root of the repo - but open to other suggestions.
</blockquote></details>

So I initially had a relative reference, but it didn't work with bazel run. At BDI we just had an environment variable for our workspace. I'll think a little more on this.

ShaneRozenLevy (commented):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324)* status: 7 of 22 files reviewed, 7 unresolved discussions (waiting on @yangwill)

___
*[`examples/Cassie/cassie_utils.cc` line 244 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzI6Cq4J75YDDuI1GH:-NG2FKsAE3abyWUenIr1:b-23kktn) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/cassie_utils.cc#L244)):*

We've separate the joints into symmetric vs asymmetric in order to correctly implement periodicity constraints. Given that we're just pushing them all to the same vector, we might want to either not distinguish between the two types OR return them as separate vectors that users can then concatenate.
</blockquote></details>

I cleaned this function up
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc` line 50 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzInqA1-4cVU8VaAjU:-NG2LjK491TVeiBNbtsV:bb7f2mf) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc#L50)):*

I like that you've moved the settings into a yaml, can we also move these gait definitions into separate yamls? Kind of like a stand/walk yaml
</blockquote></details>

done
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc` line 121 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzJF0jBretK4gsvS_k:-NG2PR9a80ZM9_9m3FZ8:bgyrfzc) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kcmpc_trajopt.cc#L121)):*
<details><summary><i>Previously, ShaneRozenLevy wrote…</i></summary><blockquote>

So I initially had a relative reference, but it didn't work with bazel run. At BDI we just had an environment variable for our workspace. I'll think a little more on this. 
</blockquote></details>

Ok so the solution I found is to use a relative path inside a try catch statement. For now most of the time I don't care too much about saving the trajectory and when I do I can just directly run the binary rather than using bazel run
___
*[`examples/Cassie/kinematic_centroidal_mpc/cassie_kinematic_centroidal_mpc.h` line 4 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzK59m4-ogYm4svDKY:-NG294enB-I17M2niCmb:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/examples/Cassie/kinematic_centroidal_mpc/cassie_kinematic_centroidal_mpc.h#L4)):*

nitpick, clion sometimes does this when it suggests imports but the correct path should be #include <drake/multibody/plant/multibody_plant.h>
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.h` line 33 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzKjMc6lHaDBtBev9y:-NG2ALxF3LXOPokRwPrR:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.h#L33)):*

i think drake style guide prefers `is_valid()`.
Though because this does not return anything, I think it should be `check_valid()`.
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.h` line 43 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzL2NE6Cbj2y5m883j:-NG2AlSw46HH4ooG8K8z:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.h#L43)):*

Also weird style preference of drake. I think trivial getters are `lower_case()` and "verbs" or functions that compute things are `GetUpperCase()`.
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/gait.cc` line 7 at r1](https://reviewable.io/reviews/dairlab/dairlib/324#-NFzLsm6CqxHKZpnuhh8:-NG2AmLEDhdtA6ZBWPDN:bjbpwmy) ([raw file](https://github.com/dairlab/dairlib/blob/39f7413baed202726e08a6c6cd158a797620c2bb/systems/controllers/kinematic_centroidal_mpc/gait.cc#L7)):*

If we're returning immediately when we find a match, and if mode.start_phase is monotonically increasing, we should only have to check `phase_now > mode.start_phase`. Though considering the second condition (which should always be true if the first condition is true in the order that we check) is after, this optimization is negligible.
</blockquote></details>

The second check is actually necessary in cases where `phase_now` is at a mode switch.

yangwill (changes requested):
Still have the main mpc class left to review

> Reviewed 8 of 22 files at r1, 11 of 11 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324#-:-NG3aFSN31IVSJcLo1O0:b-koml1z)* status: 23 of 25 files reviewed, 12 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[`systems/controllers/kinematic_centroidal_mpc/gait.cc` line 21 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3WHeG2G5AsR4gf2AL:-NG3WHeG2G5AsR4gf2AM:bwri3ei) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/gait.cc#L21)):*
> ```cpp
>   // Calculate initial mode index, and phase
>   int current_mode = GetCurrentMode(current_time);
>   double current_phase = fmod(current_time/period, 1);
> ```

took me a while, but i think i understand the class now - I think using a fixed period and the current_time could cause us issues in the future. Should be fine for now, but might want add a TODO to make this more flexible in the future. 

The case I can think of is that currently we don't have a way to stitch `Gait`s with different periods.
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.h` line 62 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3X4bWBfqSO7Pu6MXG:-NG3X4bWBfqSO7Pu6MXH:bkgmfou) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.h#L62)):*
> ```objc
> 
>  private:
>   const drake::multibody::MultibodyPlant<double>& plant_;
> ```

weird assortment of member variables, i guess this class is accomplishing many functions. Still might be worth cleaning up in the future
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc` line 54 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3RQ5G9mXzkdtw72mC:-NG3RQ5G9mXzkdtw72mD:b-r7qf44) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc#L54)):*
> ```cpp
>     dairlib::multibody::SetVelocitiesIfNew<double>(plant, v_traj.value(time), context.get());
>     auto& knot_point_value = samples.emplace_back(Eigen::VectorXd::Zero(6 * n_contact));
>     for(int i = 0; i <n_contact; i++){
> ```

not sure how this function works, the first 3 rows are the position of the contact point in the world frame with the z_height set to 0? And then last three rows are the part of the acceleration of the contact world? It's missing the `J vdot` portion of the acceleration. Just generally confused about how this function operates as the reference
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc` line 81 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3TYoZ-AxH5jPmmL9V:-NG3TYoZ-AxH5jPmmL9W:bpsdslk) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc#L81)):*
> ```cpp
> 
> void KcmpcReferenceGenerator::Build(const Eigen::Vector3d& com) {
>   com_trajectory_ = GenerateComTrajectory(com,
> ```

Is there a reason why all these functions are not member functions? Everything being passed through are member variables already. Or maybe move the rest of the reference generators into `reference_generation_utils`
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 12 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3XeOAA7DxSp1_YGbW:-NG3XeOAA7DxSp1_YGbX:ba78max) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L12)):*
> ```cpp
>                                                               double dt,
>                                                               int knot_index): dairlib::solvers::NonlinearConstraint<T>(
>     6,  2 * 6 + 2 * (3 + 2 * 3 * n_contact),
> ```

we're gonna need some justification for this arithmetic.
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 25 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3Z2h7-zxF0x6Ps2bC:-NG3Y8pE8P0SmiauPpEE:bsy9oa9) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L25)):*
> ```cpp
> 
> /// The format of the input to the eval() function is in the order
> ///   - xMom0, momentum state at time k
> ```

nitpick, feel free to leave it but these names don't make a lot of sense to me. can we use `h` for momentum? I don't understand why contact locations are `cj_`, can they be `x_contact` or `c_loc`? Also for forces, vectors should be lowercase, so `f1` or `l1` to represent abbreviate lambda.
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 164 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_NMqEfwPoFT3rHZ-:-NG3_NMqEfwPoFT3rHZ0:bi6u5xc) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L164)):*
> ```cpp
>                                                            drake::VectorX<T> *y) const {
>   const auto& x0 = x.head(n_x_);
>   const auto& r = x.segment(n_x_, 3);
> ```

making a choice here to calculate the angular momentum about the pelvis instead of the CoM.BTW, this doesn't match the comment
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 422 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3N6DI8K3O_Mq1LvDn:-NG3N6DI8K3O_Mq1LvDo:b6swdw2) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L422)):*
> ```cpp
> }
> 
> void KinematicCentroidalMPC::SetRobotStateGuess(const drake::trajectories::PiecewisePolynomial<double>& q_traj,
> ```

Can you add a TODO that will remove this function once Drake adds a trajectory stacking function? I know it's in the works and after it's implemented we can just ask the user to stack the q and v trajectories before calling the above function
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h` line 25 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3KzIp9DF6u7Nfzq2Z:-NG3KzIp9DF6u7Nfzq2_:bnfzndy) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h#L25)):*
> ```objc
>  * @param com_traj trajectory for the center of mass
>  * @param base_pos_start index where the base position starts in generalized state
>  * @return trajectory of generalized state
> ```

Few terminology questions that might help make these functions easier to understand:
- what does rt stand for?
- what does ewrt stand for?
- I think you mean generalized positions instead of generalized state - though also it seems like this function returns a trajectory in R^3, which may not be the generalized positions. Maybe floating base?
- Let's use floating base instead of base just to avoid confusion

If the spatial quantities `vel_ewrt_w` and `base_rt_com_ewrt_w` can fit nicely into the convention: https://drake.mit.edu/doxygen_cxx/group__multibody__quantities.html. Let's use that naming convention.
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc` line 22 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_utU80Vx-FlmKU6T:-NG3_utU80Vx-FlmKU6U:b-wnvul3) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc#L22)):*
> ```cpp
>                                                                                   int base_pos_start){
>   auto n_points = com_traj.get_segment_times().size();
>   std::vector<drake::MatrixX<double>> samples(n_points);
> ```

This should be a VectorXd, which also allows you to use `segment` instead of `block`
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc` line 38 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3a6JsBD8ZWthuK0Qw:-NG3a6JsBD8ZWthuK0Qx:bvxujcv) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc#L38)):*
> ```cpp
>     samples[i] = drake::VectorX<double>::Zero(n_v);
>     samples[i].block<3,1>(base_vel_start,0, 3, 1) = com_traj.derivative().value(com_traj.get_segment_times()[i]);
>   }
> ```

can use VectorXd here as well.

yangwill (changes requested):
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324)* status: 23 of 25 files reviewed, 14 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 132 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG4M7Ne2-Pk38ztAvmX:-NG4M7Ne2-Pk38ztAvmY:b-pli27q) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L132)):*
> ```cpp
>         // Feet are above the ground
>         double lb = 0;
>         if(knot_point > 0 and knot_point + 1 < n_knot_points_ and (!contact_sequence_[knot_point-1][contact_index] or !contact_sequence_[knot_point+1][contact_index])){
> ```

need a comment here: this is checking if the contact_mode before and after this mode are in flight right?
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 220 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG4Mqbw8kUV-YGL7uiD:-NG4Mqbw8kUV-YGL7uiE:b-ehonus) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L220)):*
> ```cpp
>   for (int knot_point = 0; knot_point < n_knot_points_; knot_point++) {
>     const double terminal_gain = knot_point == n_knot_points_-1 ? 100 : 1;
>     const double collocation_gain = (knot_point == 0 or knot_point == n_knot_points_ -1) ? 0.5 : 1;
> ```

We almost need `is_first_knot()` and `is_last_knot()` functions, there's been a couple time when we ask that and it would improve readability.

yangwill (commented):
Finished skimming it over - only thing that needs more attention is the CentroidalDynamicsConstraint, i think there might be a bug where the spatial momentum is calculated wrt floating base, where other parts use the center of mass.

> Reviewed 2 of 22 files at r1.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324#-:-NG4NOjg79E0XT4qthFQ:brjt9v)* status: all files reviewed, 13 unresolved discussions (waiting on @ShaneRozenLevy)

ShaneRozenLevy (commented):
Thanks for all the feedback. I made most of the changes. I had a quick question about the drake notation for spatial algebra. 

Additionally regarding the bug in the `CentroidalMomentumConstraint`, I believe its correct since the decision variable thats passed in is the center of mass, not the location of the floating base.

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324#-:-NGFFPU93g0XiO4yI98O:b46w0wi)* status: 19 of 25 files reviewed, 13 unresolved discussions (waiting on @yangwill)

___
*[`systems/controllers/kinematic_centroidal_mpc/gait.cc` line 21 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3WHeG2G5AsR4gf2AL:-NG4Ovd46WxyQQpJCyfb:blnm3tk) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/gait.cc#L21)):*

took me a while, but i think i understand the class now - I think using a fixed period and the current_time could cause us issues in the future. Should be fine for now, but might want add a TODO to make this more flexible in the future. 

The case I can think of is that currently we don't have a way to stitch `Gait`s with different periods.
</blockquote></details>

I do think there is room to phase shift gaits so the mode sequence line up. There is some question about that messing up the contact timing. I'll add a comment mentioning that switching gaits might be wonky.
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.h` line 62 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3X4bWBfqSO7Pu6MXG:-NGDBPJhCqLnIbOuMnt7:b-gq8mio) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.h#L62)):*

weird assortment of member variables, i guess this class is accomplishing many functions. Still might be worth cleaning up in the future
</blockquote></details>

Yeah I'm open to suggestions on how to clean it up. I tried to use the member variables for things that might stay constant between mpc iterations.
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc` line 54 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3RQ5G9mXzkdtw72mC:-NG4M4Pi3M3wjSy3lyyc:b-6bic6m) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc#L54)):*

not sure how this function works, the first 3 rows are the position of the contact point in the world frame with the z_height set to 0? And then last three rows are the part of the acceleration of the contact world? It's missing the `J vdot` portion of the acceleration. Just generally confused about how this function operates as the reference
</blockquote></details>

Good catch. The second 3 rows are supposed to be the velocity of the contact in the world. I think I read `EvalFullJacobianDotTimesV` as `EvalFullJacobianTimesDotq`. I'll find the correct function to use.
___
*[`systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc` line 81 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3TYoZ-AxH5jPmmL9V:-NG4MalD75r4u6ZQ-RQ6:b72k9ly) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kcmpc_reference_generator.cc#L81)):*

Is there a reason why all these functions are not member functions? Everything being passed through are member variables already. Or maybe move the rest of the reference generators into `reference_generation_utils`
</blockquote></details>

Woops. These were supposed to be move into `reference_generation_utils`. Thanks for catching this
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 12 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3XeOAA7DxSp1_YGbW:-NGCzxHA3UcRktywOpXA:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L12)):*

we're gonna need some justification for this arithmetic.
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 25 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3Z2h7-zxF0x6Ps2bC:-NGD-xRmBvsD3XcUFm0p:b-isxala) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L25)):*

nitpick, feel free to leave it but these names don't make a lot of sense to me. can we use `h` for momentum? I don't understand why contact locations are `cj_`, can they be `x_contact` or `c_loc`? Also for forces, vectors should be lowercase, so `f1` or `l1` to represent abbreviate lambda.
</blockquote></details>

done. I don't like `h` for momentum since h is traditionally used as time step duration in optimization.
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 164 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_NMqEfwPoFT3rHZ-:-NG4OD_sCGRogbYdVSqj:byikwkk) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L164)):*

making a choice here to calculate the angular momentum about the pelvis instead of the CoM.BTW, this doesn't match the comment
</blockquote></details>

It should be angular momentum about the center of mass. `r` is the location of the com, and its set equal to the com decision vars on line 147 of `kinematic_centroidal_mpc.cc` when I add the `CentroidalMomentumConstraint` to the optimization problem
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 132 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG4M7Ne2-Pk38ztAvmX:-NGD0eFt7P2e_qu8dPGr:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L132)):*

need a comment here: this is checking if the contact_mode before and after this mode are in flight right?
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 220 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG4Mqbw8kUV-YGL7uiD:-NGFFNgC2dk8_DRoScM2:bb7f2mf) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L220)):*

We almost need `is_first_knot()` and `is_last_knot()` functions, there's been a couple time when we ask that and it would improve readability.
</blockquote></details>

done
___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc` line 422 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3N6DI8K3O_Mq1LvDn:-NGD0v5D5SVynRHpjROb:b-896fix) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_mpc.cc#L422)):*

Can you add a TODO that will remove this function once Drake adds a trajectory stacking function? I know it's in the works and after it's implemented we can just ask the user to stack the q and v trajectories before calling the above function
</blockquote></details>

Done.
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h` line 25 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3KzIp9DF6u7Nfzq2Z:-NGD9mUC3py0SyiMAZoD:bjl7apg) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h#L25)):*

Few terminology questions that might help make these functions easier to understand:
- what does rt stand for?
- what does ewrt stand for?
- I think you mean generalized positions instead of generalized state - though also it seems like this function returns a trajectory in R^3, which may not be the generalized positions. Maybe floating base?
- Let's use floating base instead of base just to avoid confusion

If the spatial quantities `vel_ewrt_w` and `base_rt_com_ewrt_w` can fit nicely into the convention: https://drake.mit.edu/doxygen_cxx/group__multibody__quantities.html. Let's use that naming convention.
</blockquote></details>

rt is relative to, ewrt is expressed with regards to. These are pieces from the BD spatial quantities naming convention which I found quite good and adopted into my own code. I'll work on converting these to drake. Am I correct in my understanding of the drake notation that the vector from com to the floating base in the world frame would be `p_ComBase_World`
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc` line 22 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_utU80Vx-FlmKU6T:-NGD1v7M9fdzMxFf3W8O:bex7prb) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc#L22)):*

This should be a VectorXd, which also allows you to use `segment` instead of `block`
</blockquote></details>

Unfortunately c++ isn't able to figure out the cast here from `std::vector<Eigen::VectorXd>>` to `std::vector<Eigen::MatrixXd>>` when creating the splines.
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc` line 38 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3a6JsBD8ZWthuK0Qw:-NGDBNb30Rc-vAZfPY8y:b-34vmyd) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc#L38)):*

can use VectorXd here as well.
</blockquote></details>

See above

yangwill (approved):
> Reviewed 6 of 6 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/324)* status: all files reviewed, 3 unresolved discussions (waiting on @ShaneRozenLevy)

___
*[`systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc` line 164 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_NMqEfwPoFT3rHZ-:-NGJC8e3259q2zP-tNU9:b-xah3ci) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/kinematic_centroidal_constraints.cc#L164)):*
<details><summary><i>Previously, ShaneRozenLevy wrote…</i></summary><blockquote>

It should be angular momentum about the center of mass. `r` is the location of the com, and its set equal to the com decision vars on line 147 of `kinematic_centroidal_mpc.cc` when I add the `CentroidalMomentumConstraint` to the optimization problem
</blockquote></details>

Ah yep I missed the center of mass constraint on `com` .
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h` line 25 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3KzIp9DF6u7Nfzq2Z:-NGJ7w9g7t8XXJWv4VO5:b-2joh4o) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.h#L25)):*
<details><summary><i>Previously, ShaneRozenLevy wrote…</i></summary><blockquote>

rt is relative to, ewrt is expressed with regards to. These are pieces from the BD spatial quantities naming convention which I found quite good and adopted into my own code. I'll work on converting these to drake. Am I correct in my understanding of the drake notation that the vector from com to the floating base in the world frame would be `p_ComBase_World`
</blockquote></details>

yea `base_rt_com_ewrt_w` would be `p_ScmBase_W`, adopting the same convention for center of mass and world frame in https://github.com/RobotLocomotion/drake/blob/285452ad162100886322324e594dbb7fc83cdb91/multibody/plant/multibody_plant.h#L2762. 
___
*[`systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc` line 22 at r2](https://reviewable.io/reviews/dairlab/dairlib/324#-NG3_utU80Vx-FlmKU6T:-NGJAVlc0UqcNSwwzVyq:bf645cj) ([raw file](https://github.com/dairlab/dairlib/blob/e7f0843787c813aca34fb561c1ed0858056497e8/systems/controllers/kinematic_centroidal_mpc/reference_generation_utils.cc#L22)):*
<details><summary><i>Previously, ShaneRozenLevy wrote…</i></summary><blockquote>

Unfortunately c++ isn't able to figure out the cast here from `std::vector<Eigen::VectorXd>>` to `std::vector<Eigen::MatrixXd>>` when creating the splines.
</blockquote></details>


Ah i see, yeah that's annoying

PR #326, opened 2022-11-04
removing attic

yangwill:
We haven't needed/supported attic in years. We should remove it to prevent confusion


yminchen (approved):
Maybe ask in slack to make sure no one uses attic, and merge later

PR #325, opened 2022-11-04
Support OSC owning tracking datas as well as other QoL improvements

yangwill:
Several significant changes to OSC
- OSC now owns tracking datas, this prevents segfaults if we want to use the OSC outside of a main executable (ie. make an OSC factory)
- OSC cost logging now uses the actual drake evaluators instead of manually recomputing the cost. This prevents accidental mismatches if we change the cost formulation
- Improved impact-invariant formulation
- Add OSC gains base class to contain the common cost matrices

- - -
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/325)

PR #330, opened 2022-11-22
Fix a bug in osc tracking osc logging

yminchen:
The osc tracking cost order is inconsistent with the name's, because we save the names using `vector.push_back` while saving the cost value using indices

- - -
This change is [Reviewable](https://reviewable.io/reviews/dairlab/dairlib/330)

### DIFF
--- a/systems/controllers/osc/operational_space_control.cc
+++ b/systems/controllers/osc/operational_space_control.cc
@@ -1,12 +1,12 @@
 #include "systems/controllers/osc/operational_space_control.h"
 
-#include <drake/multibody/plant/multibody_plant.h>
 #include <iostream>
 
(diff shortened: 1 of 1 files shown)


yangwill (commented):
My osc refactor #325 might also fix this

> *[Reviewable](https://reviewable.io/reviews/dairlab/dairlib/330#-:-NHVF-Pr09XQRSOjcwdi:bg4jfay)* status: 0 of 1 files reviewed, all discussions resolved

yminchen:
> 

`push_back` + indexing caused issue when there are inactive tracking data. That is, `push_back` skips inactive tracking data but indexing does not.

Yes #325 fixes this issue. I'm closing this PR

PR #335, opened 2023-01-23
seeking to first known message to set as start time

yangwill:
Previously we were setting the timestamp of ANY lcm message as the `start_time` for the log. This works fine if we don't need precise seeking for hardware logs as we start the logger around when we start the experiment. However, for sim logs when we're often inadvertently logging visualization messages, this makes it difficult to accurately seek to the actual start time of the experiment. 

This makes a simple change to have the `start_time` line up with the first KNOWN lcm message of the log as specified by `lcm_channels`

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/335)

### DIFF
--- a/bindings/pydairlib/analysis/process_lcm_log.py
+++ b/bindings/pydairlib/analysis/process_lcm_log.py
@@ -16,6 +16,8 @@ def get_log_data(lcm_log, lcm_channels, start_time, duration, data_processing_ca
     data_to_process = {}
     print('Processing LCM log (this may take a while)...')
     lcm_log.seek(0)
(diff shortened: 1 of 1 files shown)


Brian-Acosta (approved):
lgtm

> Reviewed 1 of 1 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/335#-:-NMVbuTU9pxnvrG6f7fc:bb7k0qh)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @yangwill)

PR #359, opened 2024-05-03
updating to drake v1.28.0

yangwill:
Updating main to work with drake v1.28.0

Includes minor changes to `multibody_utils` to support model instances and `multipose_visualizer` to support meshcat.

Also cleaning out very likely unused files such as `director/` and `examples/kuka_iiwa_arm`

PR #360, opened 2024-05-03
Update readme

yangwill:
Finally updating the main README to have simpler installation instructions and moving the configuration specific instructions to a separate file.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/360)


yangwill:
Builds off of #359, so waiting for that PR to succeed first

Brian-Acosta (commented):
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/360)* status: 0 of 88 files reviewed, 1 unresolved discussion (waiting on @mposa and @yangwill)

___
*[`README.md` line 44 at r2](https://reviewable.io/reviews/DAIRLab/dairlib/360#-Nyf6ehJ61sEF70F-pvC:-Nyf6ehKARYR4gsquMJT:b-bl097k) ([raw file](https://github.com/DAIRLab/dairlib/blob/3673c977855ff661441f2b41b5c13c4e2cbe004c/README.md#L44)):*
> ```Markdown
> 
> ### Notes for macOS
> 1. Be sure to have Xcode 9.0 or later installed with Command Line Tools. If you receive a `clang: error: cannot specify -o when generating multiple output files` message during the build process, re-run `install_prereqs.sh`, and be sure that it runs fully before termination, as this will reconfigure Xcode to work with Drake.
> ```

I think this is pretty out of date. I would just link to the drake mac install instructions here, with a note that we don't test builds on mac, so ymmv. Maybe request that changes to build on mac get a PR.

examples/Cassie/osc/high_level_command.cc
Brian-Acosta, line 1: This update is basically free, let's not make it forced b/c we might want to use this system outside of LcmDrivenLoop

install/README.md
Brian-Acosta, line 1: If we are going to keep ROS Noetic instructions and infrastructure, we should tell users to revert their drake version to one that  supports Ubuntu 20.04

Brian-Acosta (commented):
A few nits otherwise LGTM

Brian-Acosta (approved):

PR #366, opened 2024-10-03
Update LcmDrivenLoop to add a message queue size parameter

Brian-Acosta:
Addresses https://github.com/RobotLocomotion/drake/issues/15234 by adding the ability to set the queue size of LcmDrivenLoop's internal subscribers, and handling the lcm subscription until the queue is empty (up to date) on each loop.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/366)

### DIFF
--- a/systems/framework/lcm_driven_loop.h
+++ b/systems/framework/lcm_driven_loop.h
@@ -60,6 +60,9 @@ namespace systems {
  * for multi-threaded locking; it should only be used in cases where the
  * governing DrakeLcmInterface::HandleSubscriptions is called from the same
  * thread that owns all copies of this object.
+ *
+ * This is a duplicate of drake::lcm::Subscriber, but makes the subscription_
+ * object public, which we need to change the queue size.
  */
 template <typename Message>
 class Subscriber final {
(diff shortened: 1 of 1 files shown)


systems/framework/lcm_driven_loop.h
```diff
@@ -157,6 +234,14 @@ class LcmDrivenLoop {
     return simulator_->get_mutable_context();
   }
 
+
+  InputMessageType wait_for_message() const {
```
mposa: Unless I missed it, you're not actually using this new method?

mposa (changes requested):
I added a bit of documentation, and just have one additional comment/question.

systems/framework/lcm_driven_loop.h
Brian-Acosta: Ah. This is used in the branch I prototyped on, not needed for a minimal fix. It can be removed

mposa (commented):
> Reviewed 1 of 1 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/366)* status: all files reviewed, 1 unresolved discussion (waiting on @sharanyashastry)

PR #371, opened 2025-05-16
Expose Franka systems for use in sampling_c3 examples

sharanyashastry:
This PR copies and exposes files from `examples/franka/systems` so that they can be reused by the sampling_c3 examples.

To build the changes introduced in this PR, run:
 `bazel build ...`

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/371)


WORKSPACE
ebianchi, line 1: If you make this PR to go from `sampling_based_c3_PR3` into `sampling_based_c3_PR1` (instead of into `_public`), then it will already have the Drake upgrade changes and not show this file in the diff.

install/install_prereqs_jammy.sh
ebianchi, line 1: Why does this file show up in this diff?

systems/end_effector_orientation.cc
ebianchi, line 1: This comment applies to all of these files:  this PR proposes _duplicating_ these files, not moving them.  E.g. this file exists at both `examples/franka/systems/end_effector_orientation.cc` and now also in `systems/end_effector_orientation.cc`.  There needs to be just one of these files, and the point of this PR is to move them to the `systems/` folder so they're accessible to all examples.

systems/robot_lcm_systems.cc
```diff
@@ -1,3 +1,4 @@
+#include <iostream>
```
ebianchi: Unnecessary.

ebianchi (changes requested):
Some comments for redirecting the approach of the PR to build off other PRs and to not duplicate code.

ebianchi:
There are also other files that are not included in this PR but that are just duplicated from the `examples/franka` folder into the next PR's `examples/sampling_c3` folder.  None of these PRs should propose duplicating entire files.  Here are some of the files I'm referring to:

- Ones that, AFAIK, should be identical across examples:
    - franka_bridge_driver_in.cc
    - franka_bridge_driver_out.cc
    - franka_joint_osc_controller.cc
    - franka_osc_controller.cc
- Some others I'm not entirely sure if there are changes or not between the two examples, but whose changes need to be justified and will be hard to identify if the entire file is repeated:
    - franka_sim.cc
    - franka_state_translator.cc
    - franka_state_translator.h

If you agree these are reused across examples, they should be included in this PR which exposes the files needed by multiple examples.

systems/end_effector_orientation.cc
sharanyashastry: Addressed this - they've all been moved under `systems/controllers/osc`

sharanyashastry:
Following files have been moved out from being exclusive to the `franka` example:

- end_effector_trajectory.cc/h
- end_effector_orientation.cc/h
- end_effector_force_trajectory.cc/h
- franka_state_translator.cc/h
- franka_kinematics.cc/h 
- c3_state_sender.cc/h
- franka_kinematics_vector.cc/h

As discussed, files with main loops have been contained within the example folders (one for `franka` and one for all `sampling_c3` examples). These include:

- franka_bridge_driver_in.cc
- franka_bridge_driver_out.cc
- franka_joint_osc_controller.cc
- franka_osc_controller.cc
- franka_sim.cc

examples/franka/diagrams/franka_osc_controller_diagram.cc
```diff
@@ -116,8 +116,10 @@ FrankaOSCControllerDiagram::FrankaOSCControllerDiagram(
   auto osc_command_sender =
       builder.AddSystem<systems::RobotCommandSender>(*plant_);
   auto end_effector_trajectory =
-      builder.AddSystem<EndEffectorTrajectoryGenerator>(
-          controller_params.neutral_position);
+      builder.AddSystem<EndEffectorTrajectoryGenerator>(*plant_, plant_context_.get(), 
+                                                        controller_params.neutral_position,
+                                                        false,
+                                                        controller_params.end_effector_name);
```
ebianchi: Make this formatting match.

examples/franka/franka_osc_controller.cc
```diff
@@ -138,7 +138,10 @@ int DoMain(int argc, char* argv[]) {
   auto osc_command_sender =
       builder.AddSystem<systems::RobotCommandSender>(plant);
   auto end_effector_trajectory =
-      builder.AddSystem<EndEffectorTrajectoryGenerator>(controller_params.neutral_position);
+      builder.AddSystem<EndEffectorTrajectoryGenerator>(plant, plant_context.get(), 
+                                                        controller_params.neutral_position,
+                                                        false,
+                                                        controller_params.end_effector_name);
```
ebianchi: Make this formatting match.

systems/controllers/osc/end_effector_force.cc
```diff
@@ -1,3 +1,4 @@
+/* This file was moved from examples/franka/systems to be used by other examples.*/
```
ebianchi: I'm unsure if this comment should be here.  Users of this code don't need to know where the file used to be located, so maybe it's better suited just as a comment in this PR discussion rather than as a comment in a file that will exist for a long time.  (Same applies to other files with this comment.)

systems/controllers/osc/end_effector_position.cc
```diff
@@ -0,0 +1,156 @@
+/**This system is used by the osc diagram to read an end effector trajectory from an lcm message and pass it onto the OSC class.
+* This system was moved from examples/franka/systems and modified to allow for teleop to track the current end effector position 
+* instead of a neutral position when needed. */
```
ebianchi: The first comment is helpful, but should be reformatted to use the same wrap column as the rest of the code (80 columns).  The second comment's history of the file movement (like the others pointed out in this PR) should get removed, but mentioning the teleop ability is useful.  I'd recommend combining the first comment with the teleop ability, and formatting the wrapping to match the rest of the file.

systems/franka_kinematics.cc
```diff
@@ -1,4 +1,6 @@
-#include "examples/franka/systems/franka_kinematics.h"
+/** This system is used to generate the lcs state used by the sampling c3 controller and was moved from
+ * examples/franka/systems. */
```
ebianchi: Remove (like others).  The mention of "sampling c3 controller" isn't relevant yet, and the file can be (and is) used by other examples so it doesn't need the callout.

systems/franka_kinematics_vector.h
```diff
@@ -41,11 +43,10 @@ class FrankaKinematicsVector : public TimestampedVector<T> {
...
   /// Constructs a OutputVector with the specified positions and velocities.
-  explicit FrankaKinematicsVector(
-      const drake::VectorX<T>& end_effector_positions,
-      const drake::VectorX<T>& object_positions,
-      const drake::VectorX<T>& end_effector_velocities,
-      const drake::VectorX<T>& object_velocities)
+  explicit FrankaKinematicsVector(const drake::VectorX<T>& end_effector_positions,
+                                  const drake::VectorX<T>& object_positions,
+                                  const drake::VectorX<T>& end_effector_velocities,
+                                  const drake::VectorX<T>& object_velocities)
```
ebianchi: This reformatting is unnecessary.  The whole file can get reverted (but stay in its new location), as the comment should be deleted as well.

systems/sender_systems/c3_state_sender.cc
```diff
@@ -14,6 +19,9 @@ C3StateSender::C3StateSender(int state_size,
   this->set_name("c3_state_sender");
 
   n_x_ = state_size;
+  final_target_state_ = this->DeclareVectorInputPort("final_target_state",
+                                               BasicVector<double>(state_size))
+                      .get_index();
```
ebianchi: Is this formatting correct?  Line things up so they match other lines, or if it crosses the 80 column width mark, wrap it a different way.

systems/sender_systems/c3_state_sender.cc
```diff
@@ -1,4 +1,9 @@
-#include "examples/franka/systems/c3_state_sender.h"
+/** This system doesn't process the data it receives in any way except for converting it into an lcm message. 
+ * This system is required since we use the generic multiplexer to concatenate the pieces of the c3 state and need to send 
+ * it out over lcm. This was moved from examples/franka/systems and modified to add an additional input and output port 
+ * to allow for a final c3_target as used by sampling_c3 examples. */ 
+
```
ebianchi: Remove.

systems/sender_systems/c3_state_sender.cc
ebianchi, line 1: It could be useful to document what the difference between `target` and `final_target` is (and at that rate, write what `actual` is too).  Could either go in a comment at the top of the file, or by each of the input port definitions, or by each of the output port functions.

ebianchi (changes requested):
Comments left.  Why did radio-related lines change?

sharanyashastry:
Regarding the radio-related changes, the code used to be structured the way I have it now, but I think @yangwill  made updates after we branched off of his version. Specifically, we’re now using an `AbstractInputPort` instead of a `VectorInputPort`. I’m not entirely sure why Will updated it, or if `VectorInputPort` is somehow better. The reason I reverted it was that the `VectorInputPort` requires a size to be hard coded, and it also requires an additional `RadioToVector` system, which I thought we could avoid.

That is, `lcmt_radio` type was used to subscribe to the radio message:
https://github.com/DAIRLab/dairlib/blob/5d86320ca001e9596fb548b113c48a6eac0d46f5/examples/franka/franka_osc_controller.cc#L151
At some point this was added:
https://github.com/DAIRLab/dairlib/blob/5d86320ca001e9596fb548b113c48a6eac0d46f5/examples/franka/franka_osc_controller.cc#L154
which converted the radio message to a Vector value and ports like these
https://github.com/DAIRLab/dairlib/blob/5d86320ca001e9596fb548b113c48a6eac0d46f5/examples/franka/systems/end_effector_position.cc#L30
needed to be declared `VectorInputPorts`.

I am proposing we remove this:
https://github.com/DAIRLab/dairlib/blob/5d86320ca001e9596fb548b113c48a6eac0d46f5/examples/franka/franka_osc_controller.cc#L154
and instead read the message from the subscriber using:
https://github.com/DAIRLab/dairlib/blob/3cbdc6eab7ba8a247b01eb681057386816228787/systems/controllers/osc/end_effector_position.cc#L44

systems/sender_systems/c3_state_sender.cc
ebianchi, line 1: I ensured these comments for differentiating C3 actual, final target, and target are more general so they apply to other C3 use cases.

ebianchi (approved):
All LGTM.

PR #369, opened 2025-05-15
Drake Version Update to Enable Building of Branch

sharanyashastry:
Upgraded Drake from version 1.22.0 to 1.33.0. This is the latest version that built without needing any additional changes.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/369)

PR #372, opened 2025-05-17
Quaternion error hessian

ebianchi:
This PR introduces the source code for computing the Hessian of the scalar angle squared between two sets of quaternions.  The derivation comes from [this script](https://github.com/ebianchi/ci_mpc_utils/blob/main/quat_hessian_compute.py), though the PR only includes the C++ implementation of the resulting expressions.

This contribution can stand alone as a smaller PR to remove weight from other chunky PRs for the sampling-based C3 project, though this set of changes can be more broadly useful.

I will mark this PR as ready for review after `sampling_based_c3_PR1` gets merged into `sampling_based_c3_public` (i.e. after #369 is merged), since this `quat_hessian` branch requires the same Drake upgrade to properly `bazel build ...`.

Update 5/19/2025:  #369 was merged, so this is ready for review.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/372)


sharanyashastry:
Is `dairlib/systems/controllers` the right location for this? I was thinking maybe `dairlib/common`  makes more sense. We could also potentially combine with the `blending_utils.cc` and rename it `math_utils.cc` since it also just has small math functions used in other examples

ebianchi:
> Is dairlib/systems/controllers the right location for this? I was thinking maybe dairlib/common makes more sense.

Agreed, done.

sharanyashastry (approved):

PR #381, opened 2025-06-30
Fixes to bazel build files

mposa:
Likely a result of the upgrade to Bazel 8.2, which leads to errors on empty globs. I wonder if these empty globs weren't actually needed in the first place.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/381)


ebianchi (approved):
LGTM

PR #383, opened 2025-07-02
OSQP fix after Drake (and OSQP) upgrade

ebianchi:
Two changes for interacting with OSQP with upgraded Drake:

1. When using OSQP v0.6.3, Drake no longer exposes the OSQP-internal setting `linsys_solver` as it did with v0.6.2.  Given all of our examples so far have only ever used the default solver, this PR removes `linsys_solver` as an available setting in our solver yaml files.  To keep this as an available setting would require either customizing OSQP or Drake, neither of which felt worth it given we aren't intending to use the feature.
2. OSQP v0.6.3 requires `adaptive_rho_interval` to be provided as an `int` (i.e. it now throws an error if this is not the case).  Our code previously mixed up `int` and `double` inconsistently in a couple places, which eventually gave a `double` to OSQP.  This parameter now is appropriately defined as an `int` consistently.

The jack sampling C3 example did not run before (due to an OSQP error) and now does.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/383)

### DIFF
--- a/bindings/pydairlib/analysis/osqp_settings_sandbox.py
+++ b/bindings/pydairlib/analysis/osqp_settings_sandbox.py
@@ -165,7 +165,6 @@ def main():
          'eps_prim_inf': 1e-6,
          'eps_dual_inf': 1e-6,
          'alpha': 1.6,
-         'linsys_solver': 0,
          'delta': 1e-6,
          'polish': 1,
          'polish_refine_iter': 3,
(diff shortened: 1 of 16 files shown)


cuie23 (approved):
> Reviewed 12 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/383)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on @ebianchi)

PR #391, opened 2025-07-17
Fix bug in external force tracking feature of OSC

xuanhien070594:
In the `UpdateDynamics` function of the `OperationalSpaceControl` class, an empty list of external contact force names is currently passed, preventing the Jacobian matrix from being updated and resulting in improper force tracking.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/391)


ebianchi (approved):
LGTM

PR #393, opened 2025-08-08
add feature for changing force between obj-obj obj-wall to planar

haorany8:
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/393)


solvers/lcs_factory.cc
```diff
@@ -34,12 +34,26 @@ LCS LCSFactory::LinearizePlantToLCS(
...
+    if (IsObjToObjContact(plant,context,pair) && (!with_z_lambda)) {
+      num_planar_contacts++;
+      num_direction_contacts_vector.push_back(1);
+    }else {
+      num_direction_contacts_vector.push_back(num_friction_directions);
+    }
+  }
+
+
+  int num_direction_contacts = with_z_lambda? 2 * n_contacts * num_friction_directions : 2*num_friction_directions*(n_contacts - num_planar_contacts) + 2 * 1 * num_planar_contacts;
```
ebianchi: Recommend renaming `n_lambda`.

solvers/lcs_factory.cc
```diff
@@ -281,62 +301,104 @@ LCS LCSFactory::LinearizePlantToLCS(
...
   MatrixXd J_n(n_contacts, n_v);
-  MatrixXd J_t(2 * n_contacts * num_friction_directions, n_v);
+  MatrixXd J_t(num_direction_contacts, n_v);
   std::vector<VectorXd> contact_points;
+
   for (int i = 0; i < n_contacts; i++) {
-    multibody::GeomGeomCollider collider(plant, contact_geoms[i]);
-    auto [phi_i, J_i] = collider.EvalPolytope(context, num_friction_directions);
-    auto [p_WCa, p_WCb] = collider.CalcWitnessPoints(context);
-    // TODO(yangwill): think about if we want to push back both witness points
```
ebianchi: If this is still relevant, leave it in.

solvers/lcs_factory.cc
```diff
@@ -281,62 +301,104 @@ LCS LCSFactory::LinearizePlantToLCS(
...
+    } else {
+      auto [phi_i, J_i] =
+          collider.EvalPolytope(context, num_direction_contacts_vector[i]);
+      auto [p_WCa, p_WCb] = collider.CalcWitnessPoints(context);
+      contact_points.push_back(p_WCa);
+      phi(i) = phi_i;
+      J_n.row(i) = J_i.row(0);
+      J_t.block(2 * std::accumulate(num_direction_contacts_vector.begin(),num_direction_contacts_vector.begin()+i,0), 0, 2 * num_direction_contacts_vector[i],
+                n_v) = J_i.block(1, 0, 2 * num_direction_contacts_vector[i], n_v);
+    }
```
ebianchi: The only difference between the `if` and `else` is which normal to use.  Suggest structure like:
```
if (num_direciton_contacts_vector[i] == 1) { // Do EvalPlanar }
else { // Do EvalPolytope }
// Then after the if/else do all the other commands, starting with collider.CalcWitnessPoints
```

solvers/lcs_factory.cc
```diff
@@ -540,12 +602,45 @@ void LCSFactory::PrintVerboseContactInfo(const MultibodyPlant<double>& plant,
...
 
-  std::cout << "Contact pair: (" << inspector.GetName(pair.first()) 
+  std::cout << "Contact pair: (" << inspector.GetName(pair.first())
             << ", " << inspector.GetName(pair.second())
             << ") with phi = " << phi_i << " between world points ["
             << p_world_contact_a.transpose() << "], ["
             << p_world_contact_b.transpose() << "]" << std::endl;
 }
 
+bool LCSFactory::IsObjToObjContact(const drake::multibody::MultibodyPlant<double>& plant,
```
ebianchi: Flagging here that our in-person discussion considered making this more general purpose than detecting obj-obj filenames and instead defining a new vector that can label obj-obj interactions as "planar" (and same with obj-wall, and label e.g. ee-obj interactions as not planar, etc.).

examples/sampling_c3/anything/parameters/sampling_c3_options.yaml
```diff
@@ -14,6 +14,8 @@ publish_frequency: 0
 penalize_changes_in_u_across_solves: false  # Penalize (u-u_prev) instead of u.
 num_friction_directions: 2
 
+with_z_lambda: true
```
ebianchi: Suggest removing this parameter and instead define one like `contact_pair_types_are_planar = [false, false, false, true, true]` which would mean EE-ground, EE-obj, and obj-ground pairs are 3D and obj-obj and obj-wall pairs are planar.

ebianchi (changes requested):
See comments throughout.

PR #396, opened 2025-08-24
Fix bugs in defining object-object and object-wall contact pairs

xuanhien070594:
This PR aims to fix the three bugs below:

1. Object–object pairs: Contact pairs are currently defined over all convex decompositions of each object. As a result, when selecting three pairs among objects A, B, and C, the controller might choose, for example, two A–B pairs and one A–C pair, leaving out B–C (or even three A–B pairs, meaning object C has no modeled collisions).

2. Object–wall pairs: The definitions are broken because they previously relied on the full object mesh, while objects are now represented as multiple convex pieces.

3. include_walls option: Even if include_walls is set to false, the object–wall section of resolve_contacts_to_lists may assign a non-zero value, effectively overriding the include_walls setting.

4. We use `ang_error_sample_retention` value for filtering out outdated samples based on distance.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/396)


cuie23 (changes requested):
> @cuie23 reviewed 20 of 20 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/396)* status: all files reviewed, 2 unresolved discussions (waiting on @ebianchi)

___
*[`examples/sampling_c3/multiyaml_rewrite.py` line 138 at r1](https://reviewable.io/reviews/DAIRLab/dairlib/396#-OYSYn5IADIyenIjclZD:-OYSYn5IADIyenIjclZE:bx8w49w) ([raw file](https://github.com/DAIRLab/dairlib/blob/e11ed48c362554b5a013c9aded54c566ba1de10c/examples/sampling_c3/multiyaml_rewrite.py#L138)):*
> ```Python
>     return q_vector
> 
> def update_c3_options(is_c3_plus, samp_c3_options_yaml_path): 
> ```

Would probably also be a good idea to auto-update mu_per_pair_type since resolve_contacts_to_lists changes length depending on number of objects and/or if include wall
___
*[`examples/sampling_c3/multiyaml_rewrite.py` line 141 at r1](https://reviewable.io/reviews/DAIRLab/dairlib/396#-OYSYN3M16M4dh0-Yl-5:-OYSYN3M16M4dh0-Yl-6:b-k6wy1j) ([raw file](https://github.com/DAIRLab/dairlib/blob/e11ed48c362554b5a013c9aded54c566ba1de10c/examples/sampling_c3/multiyaml_rewrite.py#L141)):*
> ```Python
>     samp_c3_options_yaml = load_yaml(samp_c3_options_yaml_path)
> 
>     include_walls = 1 if (samp_c3_options_yaml['include_walls']) else 0
> ```

Depends on what the desired behavior is, but we probably want 2 wall contacts per object (to deal with corner

xuanhien070594:
*[`examples/sampling_c3/multiyaml_rewrite.py` line 141 at r1](https://reviewable.io/reviews/DAIRLab/dairlib/396#-OYSYN3M16M4dh0-Yl-5:-OYUEXy_-O74XPhFfVPg:b-y07g2e) ([raw file](https://github.com/DAIRLab/dairlib/blob/e11ed48c362554b5a013c9aded54c566ba1de10c/examples/sampling_c3/multiyaml_rewrite.py#L141)):*
<details><summary><i>Previously, cuie23 (Eric Cui) wrote…</i></summary><blockquote>

Depends on what the desired behavior is, but we probably want 2 wall contacts per object (to deal with corner
</blockquote></details>

The current control rate is already pretty low with long planning horizon, so I keep 1 for now, if we observe any problems with corners, we can change to 2.

xuanhien070594:
*[`examples/sampling_c3/multiyaml_rewrite.py` line 138 at r1](https://reviewable.io/reviews/DAIRLab/dairlib/396#-OYSYn5IADIyenIjclZD:-OYUEXPaCWOfNEhnJzct:b-896fix) ([raw file](https://github.com/DAIRLab/dairlib/blob/e11ed48c362554b5a013c9aded54c566ba1de10c/examples/sampling_c3/multiyaml_rewrite.py#L138)):*
<details><summary><i>Previously, cuie23 (Eric Cui) wrote…</i></summary><blockquote>

Would probably also be a good idea to auto-update mu_per_pair_type since resolve_contacts_to_lists changes length depending on number of objects and/or if include wall
</blockquote></details>

Done.

cuie23 (approved):

PDKy:
I found a bug in the MaintainSampleBuffer function in sampling_based_c3_controller.cc. https://github.com/DAIRLab/dairlib/blob/e119fb754639cf1c4f108c608010392c113f01a9/systems/controllers/sampling_based_c3_controller.cc#L1647. For this line, it should be < sampling_params.pos_error_sample_retention not <sampling_params.ang_error_sample_retention

haorany8:
https://github.com/DAIRLab/dairlib/blob/a49a6d8e0d1924fb122606fdd03fcd8c0b9c10d3/systems/controllers/sampling_based_c3_controller.cc#L1647
sampling_params_.ang_error_sample_retention should be replaced by sampling_params_.pos_error_sample_retention

xuanhien070594:
> https://github.com/DAIRLab/dairlib/blob/a49a6d8e0d1924fb122606fdd03fcd8c0b9c10d3/systems/controllers/sampling_based_c3_controller.cc#L1647
> 
> 
> sampling_params_.ang_error_sample_retention should be replaced by sampling_params_.pos_error_sample_retention

Done

haorany8:
We should add a BoundingBoxConstraint to enforce lambda and eta for the ee–object contact to be greater than zero to prevent calculating wrong direction of input. 

```
  for (int i = 0 ; i < N_; i++) {
      for (int j = 0 ; j < 4; j++) {
          prog_.AddBoundingBoxConstraint(0, std::numeric_limits<double>::infinity(),lambda_.at(i)[j]);
          prog_.AddBoundingBoxConstraint(0, std::numeric_limits<double>::infinity(),eta_.at(i)[j]);
      }
  }
```

PR #399, opened 2025-08-27
Add feature of adding non-negative constraint on lambda and eta

haorany8:
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/399)


examples/sampling_c3/multiyaml_rewrite.py
haorany8, line 1: This whitespace issue is weird. I edited the file in CLion. Even when I add those spaces, they disappear after I push.

PR #400, opened 2025-08-28
Upgrade franka_joint_osc_controller to allow moving Franka to different target joint positions

xuanhien070594:
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/400)


systems/franka_target_joint_position_receiver.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+#include "drake/systems/framework/basic_vector.h"
+#include "drake/systems/framework/leaf_system.h"
+using drake::systems::BasicVector;
+
+namespace dairlib {
+namespace systems {
+/// Receives the output of an LcmSubscriberSystem that subscribes to the
+/// robot input channel with LCM type lcmt_robot_input and outputs the
+/// robot inputs as a TimestampedVector.
```
ebianchi: The output is a `BasicVector` of size 7 for the 7 franka joints, not a `TimestampedVector`.

examples/sampling_c3/joint_trajectory_generator.cc
```diff
@@ -35,10 +34,18 @@ JointTrajectoryGenerator::JointTrajectoryGenerator(
       plant.num_velocities(), plant.num_actuators()))
     .get_index();
 
+  target_joint_position_port_ =
+    this->DeclareVectorInputPort("lcmt_franka_target_joint_position",
```
ebianchi: A bit misleading to have `lcmt_` in the name of the input port when the value is a vector input, instead of the lcmt type.

ebianchi (approved):
A couple nits, otherwise LGTM.

systems/franka_target_joint_position_receiver.h
xuanhien070594: Done

examples/sampling_c3/joint_trajectory_generator.cc
xuanhien070594: Done

PR #401, opened 2025-08-29
Keep track of previously attempted samples and avoid repeating if ineffective

ebianchi:
This change adds the ability to avoid repeating attempting near previously ineffective samples.  The "unsuccessful sample buffer" keeps track of system configurations when the controller enters C3 mode.  These configurations are cleaned out when the object(s) move beyond a threshold in position and/or orientation -- this reuses the threshold for maintaining relevant samples in the sample buffer.

The following are new hyperparameters for this feature, all in sampling parameters:

- `avoid_choosing_unsuccessful_samples` (bool) turns on this feature.  When the feature is off, the unsuccessful sample buffer is still maintained but has no effect on what samples can be pursued by the controller.
- `N_unsuccessful_sample_buffer` (int) is the size of the unsuccessful sample buffer.  In my single object tests in simulation, I only saw a maximum of 5 added to this buffer at a time, so I believe a value of 10 is sufficient.  If the unsuccessful sample buffer ever fills up, a warning print notifies the user.
- `unsuccessful_radius` (double) is the distance to any previously attempted sample within which more samples are rejected.

The unsuccessful sample buffer is published over LCM via the `UNSUCCESSFUL_SAMPLE_BUFFER` channel.

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/401)

### DIFF
--- a/examples/sampling_c3/anything/parameters/sampling_params.yaml
+++ b/examples/sampling_c3/anything/parameters/sampling_params.yaml
@@ -28,6 +28,12 @@ N_sample_buffer: 200
 pos_error_sample_retention: 0.002
 ang_error_sample_retention: 0.02  # 0.0349 rad = 2 deg
 
+# Unsuccessful sample buffer parameters.  Position and angular error sample
+# retention match that for the sample buffer above.
+avoid_choosing_unsuccessful_samples: true
+N_unsuccessful_sample_buffer: 10
+unsuccessful_radius: 0.01
+
 # Shared across multiple sampling strategies.
 sampling_radius: 0.16  # kRadiallySymmetric, kRandomOnCircle, kRandomOnSphere
 sampling_height: 0.005 # kRadiallySymmetric, kRandomOnCircle, kRandomOnPerimeter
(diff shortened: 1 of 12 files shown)


siddhantmody (changes requested):
> @siddhantmody reviewed 9 of 12 files at r1.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/401)* status: 9 of 12 files reviewed, 1 unresolved discussion

___
*[`systems/controllers/sampling_based_c3_controller.cc` line 1093 at r1](https://reviewable.io/reviews/DAIRLab/dairlib/401#-OYrEU7w4jrvjzK-i4uy:-OYrEU7w4jrvjzK-i4uz:bpxhpq9) ([raw file](https://github.com/DAIRLab/dairlib/blob/097debc0e30818ba30234be61c353097dae8db82/systems/controllers/sampling_based_c3_controller.cc#L1093)):*
> ```cpp
>       }
>       pursued_target_source_ = PursuedTargetSource::kNoTarget;
>       // Add the current state to the unsuccessful sample buffer.  It gets
> ```

Should note that it only gets automatically removed if the object moves a certain distance.

ebianchi:
> Should note that it only gets automatically removed if the object moves a certain distance.

Done.

siddhantmody (approved):

PR #406, opened 2026-04-28
feat(osc):  refactor

Meow404:
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/406)


ebianchi (approved):
Changes look good, and confirmed the Push Anything example still works as expected in simulation (which uses the OSC).  LGTM.

Issue #408, opened 2026-06-09
C3 performance issue (at least relevant to MIQP projection)


ebianchi:
The newer C3-repo-based implementation of the sampling C3 controller has a slower C3 solve than older dairlib-repo-based implementation, at least for the `jacktoy` demo using the MIQP projection.  For this setup, the newer C3 `Solve` function takes ~5x longer.  It is unclear if the performance difference impacts more sampling C3 demos and/or more C3 projection steps (e.g. C3+).

For an apples-to-apples comparison:

- [Old implementation](https://github.com/DAIRLab/dairlib/tree/bibit/sc3_with_c3_repo_timing_debug) can be run on the `bibit/sc3_with_c3_repo_timing_debug` branch (derived from `sampling_based_c3_public`)
    - Controller runs at ~10-23 Hz.
    - Parallelized C3 solve loop takes ~45-53 ms.
    - One call of `test_c3_object->Solve(test_state)` takes ~40-50ms.
- [New implementation](https://github.com/DAIRLab/dairlib/tree/bibit/debug_miqp_performance) can be run on the `bibit/debug_miqp_performance` branch (derived from `push_anything_dev`)
    - Controller runs at ~1-7 Hz.
    - Parallelized C3 solve loop takes ~230-350ms.
    - One call of `test_c3_object->Solve(test_state)` takes ~220-350ms.

These branches share the same C3 setup, including constraints and solver options.

PR #410, opened 2026-06-16
Fix other SC3 demos broken by Push Anything

ebianchi:
The original 2 sampling C3 demos included the jack (3D) and push-T (planar).  These demos now run again alongside the third SC3 demo, Push Anything (multi-object planar), after:

- fixing some code that was added for Push Anything that broke the other examples,
- retuning the original 2 demos to use the newer C3+ algorithm.  (Note:  There is some performance issue with the current implementation (in the C3 repo) of the older C3 algorithm (specifically with the MIQP projection) which is tracked here #408; this current PR does not address those issues but instead re-introduces the old SC3 examples with the new C3+ implementation).

This PR ensures experiments can be run via each of the 3 demo scripts:

- `examples/sampling_c3/jacktoy/franka_sim_jack.pmd`
- `examples/sampling_c3/push_t/franka_sim_t.pmd`
- `examples/sampling_c3/anything/franka_sim_anything.pmd`

- - -
This change is [Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/410)


xuanhien070594 (approved):
:lgtm:

> @xuanhien070594 reviewed 33 files and all commit messages, and made 1 comment.
> *[Reviewable](https://reviewable.io/reviews/DAIRLab/dairlib/410#-:-OvIWyhd9g8p8_F-NM7Z:bnfp4nl)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on ebianchi).

Meow404:
Thank you for this! examples look good. 

One nit : the visualizer for the `T` and `Jacktoy` example doesn't start on its own when you run the `start_experiment` in simulation unlike in `anything`.

Also for prioritization, is the MIQP projection in https://github.com/DAIRLab/dairlib/issues/408 still a priority item for looking into?

ebianchi:
**Visualizer:**  It's nice to start the visualizer separately and not have it restart when you start an experiment, so that you can tweak the meshcat settings inside the visualizer window once and not between every experiment.  Push Anything is the exception:  we made the visualizer restart with every Push Anything experiment because the objects can change, so the visualizer needs to reload new object models each time.  Thus I recommend we leave the procman scripts as-is.

**C3 performance issue:**  For the MIQP projection, I was able to confirm that the longest time per control loop is spent in calls to Gurobi, which handles the projection step.  Thus it doesn't seem like the issue affects C3+ examples, though it'd be best to figure this out eventually since this points to something problematic in one of the currently offered projections.  Debugging this is no longer holding me back, since I trust the C3+ formulation is unaffected.  Feel free to use that information to help prioritize.

